PR #89, opened 2013-11-22
add array coercion

motevets:
This PR adds support for array coercion/casting for spatial columns in ActiveRecord using activerecord-postgis-adapter leveraging [to_wkt_](http://rubygems.org/gems/to_wkt_).

This update allows for something like:

```ruby
foo.lonlat = [42, -42]
```

Under the hood, it's looking if the column is `:spatial` and if the value coming in, to be cast, is an array.  If it is, it converts the array into WKT and proceeds as usual.

The code has been updated for compatibility with AR3 and AR4 and is passing all of the tests there were passing previously off of your current master.


teeparham:
Closing in favor of #147

PR #119, opened 2014-05-28
Add some features for PostGis 2.1

phlegx:
- Add support for **postgis_tiger_geocoder**
- Add bug fix for **automatic schema creation**  if not exists
- Add raise error for **extension check**  if not exists


phlegx:
Close because deprecated.

PR #121, opened 2014-06-06
Update column migration method to correctly set null: false

srt32:
- Update `MainAdapter#create_table` to call `add_spatial_column`

Edit:  Here is more specific information (taken from a comment that is now deleted):

Previously running 

```
create_table :my_table do |t|
  t.column :geometry, :geometry, null: false
end
```

would produce a table with the `geometry` column without `null: false`.  With this update, including the `null: false` flag will now set the column to be `null: false` as expected.

### DIFF
--- a/lib/active_record/connection_adapters/postgis_adapter/main_adapter.rb
+++ b/lib/active_record/connection_adapters/postgis_adapter/main_adapter.rb
@@ -112,13 +112,15 @@ module ActiveRecord  # :nodoc:
             table_definition = td
           end
           table_definition.non_geographic_spatial_columns.each do |col|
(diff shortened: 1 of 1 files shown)


jeremieweldin:
Hey guys, still not sure how to use this. How would I use add_spatial_column to add a polygon column using srid 4326?

teeparham:
@jeremieweldin this is an old, closed issue that applies to the internals of the adapter. Check the readme for version 2 or 3, depending on which version of rails you're using. There are examples for adding polygon columns there.

jeremieweldin:
Sorry about that... limited results from the google search. :/  I will look into the readme. Thanks!

PR #147, opened 2014-11-07
add array coercion (updated)

ericcj:
updated version of https://github.com/rgeo/activerecord-postgis-adapter/pull/89 with master merged in


ericcj:
@teeparham not sure if this functionality is in the scope of what you want to support, but i merged master and verified it's working

BuonOmo:
@keithdoggett I'd refrain from extending ruby object with a gem that isn't updated since 2013 and has not much stars. In favor of closing. WDYT?

keithdoggett:
@BuonOmo same here. Closing.

Issue #155, opened 2014-12-22
Incompatibility with schema_plus: Error "column does not exist" on geometry column index creation


dealbase:
When using schema_plus gem and have an index defined on a non-geographic spatial column, db:schema:load errors with "column does not exist" when creating the index. Schema_plus dumps the index inline, and the schema_plus adapter tries to create the index before the column has been created by ActiveRecord::ConnectionAdapters::PostGISAdapter::MainAdapter.

For example:

```
  create_table "samples", force: true do |t|
    t.spatial  "geom",                        limit: {:srid=>4326, :type=>"point"}
    t.index ["geom"], :name => "idx_samples_geom", spatial: true
  end
```

`rake db:schema:load` will error with:

```
-- create_table("samples":force=>true})
rake aborted!
ActiveRecord::StatementInvalid: PG::UndefinedColumn: ERROR:  column "geom" does not exist
: CREATE  INDEX  "idx_samples_geom" ON "samples" USING GIST ("geom")
 /gems/activerecord-4.1.1/lib/active_record/connection_adapters/postgresql/database_statements.rb:128:in `async_exec'
 /gems/activerecord-4.1.1/lib/active_record/connection_adapters/postgresql/database_statements.rb:128:in `block in execute'
 /gems/activerecord-4.1.1/lib/active_record/connection_adapters/abstract_adapter.rb:373:in `block in log'
 /gems/activesupport-4.1.1/lib/active_support/notifications/instrumenter.rb:20:in `instrument'
 /gems/activerecord-4.1.1/lib/active_record/connection_adapters/abstract_adapter.rb:367:in `log'
  /gems/activerecord-4.1.1/lib/active_record/connection_adapters/postgresql/database_statements.rb:127:in `execute'
/work/activerecord-postgis-adapter/lib/active_record/connection_adapters/postgis_adapter/main_adapter.rb:167:in `add_index'
  /gems/activerecord-4.1.1/lib/active_record/connection_adapters/abstract/schema_statements.rb:206:in `block in create_table'
  /gems/activerecord-4.1.1/lib/active_record/connection_adapters/abstract/schema_statements.rb:206:in `each_pair'
  /gems/activerecord-4.1.1/lib/active_record/connection_adapters/abstract/schema_statements.rb:206:in `create_table'
  /gems/schema_plus-1.7.1/lib/schema_plus/active_record/connection_adapters/schema_statements.rb:38:in `create_table_with_schema_plus'
[...]

PR #156, opened 2014-12-22
fix Issue# 155 by explicitly adding non-geographic spatial indexes after...

dealbase:
... adding the columns


TylerRick:
Reference: #155

teeparham:
This sounds like a `schema_plus` problem.

PR #165, opened 2015-02-19
Renames OID::Spatial#binary to #binary_string? -- Fixes #164

### LINKED ISSUE #164: NoMethodError: private method `binary?'
Running into this error when trying to create a record with an `st_point` column. The record is still created and returned but the error still happens.

```
Could not log "sql.active_record" event. NoMethodError: private method `binary?' called for #<ActiveRecord::ConnectionAdapters::PostGISAdapter::OID::Spatial:0x007fa5cf54a4f8>
```

mrnugget:
Renaming the method solves the issue of the method being called by ActiveRecord's LogSubscriber, which expects `OID::Spatial` to inherit or overwrite the `binary?` method of ActiveRecord's `Type::Value`.

If ActiveRecord called the method before this change a NoMethodError was thrown, since the `binary?` method was private.

This change renames the method since it's for internal use only.

---

Please see #164 for more details.

What do you think? Is the name okay? Or did I misunderstand the part about `binary?(string)` being used for internal use only?

### DIFF
--- a/lib/active_record/connection_adapters/postgis_adapter/oid/spatial.rb
+++ b/lib/active_record/connection_adapters/postgis_adapter/oid/spatial.rb
@@ -100,12 +100,12 @@ module ActiveRecord
             nil
           end
 
-          def binary?(string)
+          def binary_string?(string)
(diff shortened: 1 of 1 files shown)


teeparham:
Thank you!

PR #166, opened 2015-02-23
Allow dumped schema to be reloaded

### LINKED ISSUE #161: Using `st_point` type in migration writes out a schema that cannot be reloaded.
I'm unclear to the internals of this adapter, but I think something may be off here:

In a migration:

```ruby
class CreateTrackPoints < ActiveRecord::Migration
  def change
    create_table :track_points do |t|
      t.st_point :coordinate, geographic: true
    end
  end
end
```

Executing this writes out this into my `schema.rb`:

```ruby
create_table "track_points", force: :cascade do |t|
  t.geography "coordinate", limit: {:srid=>4326, :type=>"point", :geographic=>true}
end
```

That's fine and well, but any time I attempt to run any rake command involving `db:schema:load`, I get the following error:

```
/opt/rubies/2.2.0/bin/ruby -I/opt/rubies/2.2.0/lib/ruby/gems/2.2.0/gems/rspec-core-3.1.7/lib:/opt/rubies/2.2.0/lib/ruby/gems/2.2.0/gems/rspec-support-3.1.2/lib /opt/rubies/2.2.0/
lib/ruby/gems/2.2.0/gems/rspec-core-3.1.7/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb  --exclude-pattern spec/features/\*\*/\*_spec.rb
rake aborted!
ActiveRecord::StatementInvalid: PG::InvalidParameterValue: ERROR:  Invalid geometry type modifier: geography
LINE 1: ...ck_points" ("id" serial primary key, "coordinate" geometry(G...
                                                             ^
: CREATE TABLE "track_points" ("id" serial primary key, "coordinate" geometry(GEOGRAPHY,4326), "elevation" decimal, "heart_rate" integer, "time" timestamp, "track_id" integer)
/Users/andrewhao/workspace/velocitas-core/db/schema.rb:35:in `block in <top (required)>'
/Users/andrewhao/workspace/velocitas-core/db/schema.rb:14:in `<top (required)>'
-e:1:in `<main>'
PG::InvalidParameterValue: ERROR:  Invalid geometry type modifier: geography
LINE 1: ...ck_points" ("id" serial primary key, "coordinate" geometry(G...
                                                             ^
/Users/andrewhao/workspace/velocitas-core/db/schema.rb:35:in `block in <top (required)>'
/Users/andrewhao/workspace/velocitas-core/db/schema.rb:14:in `<top (required)>'
-e:1:in `<main>'
Tasks: TOP => db:test:load_schema
```

Would it make sense to have the adapter write out the following statement instead into schema.rb?

```diff
+    t.st_point "coordinate", :geographic=>true
-    t.geography "coordinate", limit: {:srid=>4326, :type=>"point", :geographic=>true}
```

Doing so seems to fix the problem.

afn:
Fixes #161 and #167

Rather than the approach suggested in #161, I've made a small change that allows the dumped schema to be loaded.  ~~Not familiar enough with the codebase to write a proper test for this; can someone help out?~~ EDIT: I've added tests for both #161 and #167

Example schema.rb:

```ruby
  create_table "poo", id: false, force: :cascade do |t|
    t.geography "blah", limit: {:srid=>4326, :type=>"point", :geographic=>true}
  end
```

`rake db:test:prepare` previously blew up on this schema, but works correctly now.

### DIFF
--- a/lib/active_record/connection_adapters/postgis_adapter/spatial_table_definition.rb
+++ b/lib/active_record/connection_adapters/postgis_adapter/spatial_table_definition.rb
@@ -12,14 +12,15 @@ module ActiveRecord  # :nodoc:
         # super: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb#L320
         def new_column_definition(name, type, options)
           if (info = MainAdapter.spatial_column_options(type.to_sym))
+            if (limit = options.delete(:limit))
+              options.merge!(limit) if limit.is_a?(::Hash)
+            end
+
             geo_type = ColumnDefinition.geo_type(options[:type] || type || info[:type])
             base_type = info[:type] || (options[:geographic] ? :geography : :geometry)
 
             # puts name.dup << " - " << type.to_s << " - " << options.to_s << " :: " << geo_type.to_s << " - " << base_type.to_s
 
-            if (limit = options.delete(:limit))
-              options.merge!(limit) if limit.is_a?(::Hash)
-            end
             if options[:geographic]
               options[:limit] = ColumnDefinition.options_to_limit(geo_type, options)
             end
--- a/lib/active_record/connection_adapters/postgis_adapter/spatial_column.rb
+++ b/lib/active_record/connection_adapters/postgis_adapter/spatial_column.rb
@@ -44,12 +44,19 @@ module ActiveRecord  # :nodoc:
                     :geometric_type,
                     :has_m,
                     :has_z,
-                    :limit, # override
                     :srid
 
         alias :geographic? :geographic
         alias :has_z? :has_z
         alias :has_m? :has_m
+
+        def limit
+          if spatial?
+            @limit
+          else
+            super
+          end
+        end
 
         def spatial?
           cast_type.respond_to?(:spatial?) && cast_type.spatial?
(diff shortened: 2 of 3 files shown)


afn:
This now also fixes #167 (which caused limits for non-spatial columns to disappear in schema dumps)

lib/active_record/connection_adapters/postgis_adapter/spatial_table_definition.rb
```diff
@@ -12,14 +12,15 @@ def initialize(types, name, temporary, options, as, adapter)
         # super: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb#L320
         def new_column_definition(name, type, options)
           if (info = MainAdapter.spatial_column_options(type.to_sym))
+            if (limit = options.delete(:limit))
+              options.merge!(limit) if limit.is_a?(::Hash)
```
afn: @teeparham Not a huge deal, but note that this silently discards `options[:limit]` in the event that it's not a hash.  (It's deleted from `options` regardless of its type, and it's ignored if it's not a hash.)  Is this intentional?  If not, it may be better to raise an error here if `limit` is not a hash.

teeparham: It's not intentional. I agree it'd be better to raise an error.

teeparham:
Thank you for this PR.

Would you might deleting https://github.com/amitree/activerecord-postgis-adapter/commit/bab476914d8e8e8a9 (or squashing it with https://github.com/amitree/activerecord-postgis-adapter/commit/10796059c3d04e). That would clean up the commit history.

afn:
Done.

PR #170, opened 2015-04-02
Feature/smart column factories

bbradbury:
This is a stab at relieving some pain on #163 with ActiveRecord 4.2 by taking advantage of information kept with spatial columns in the database. I'm missing a bunch of broader context about how all this fits together because I've only been looking at it for a couple of days with no GIS background, but maybe something like this could provide reasonable defaults to be overridden by a `SpatialFactoryStore` as described by @teeparham and @mrnugget in #163?

### DIFF
--- a/lib/active_record/connection_adapters/postgis_adapter/oid/spatial.rb
+++ b/lib/active_record/connection_adapters/postgis_adapter/oid/spatial.rb
@@ -10,8 +10,18 @@ module ActiveRecord
           #   "geometry(Polygon,4326) NOT NULL"
           #   "geometry(Geography,4326)"
           def initialize(oid, sql_type)
+            puts "INITIALIZING: #{oid} #{sql_type}"
             @geo_type, @srid, @has_z, @has_m = self.class.parse_sql_type(sql_type)
-            @factory_generator = RGeo::Geographic.spherical_factory(srid: 4326) if oid =~ /geography/
+            puts " ==> : @geo_type:[#{@geo_type}] @srid:[#{@srid}] @has_z:[#{@has_z}]  @has_m:[#{@has_m}]"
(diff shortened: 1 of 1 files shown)


teeparham:
Thanks for working on this problem. I merged #176 instead, which is a similar solution. So I'm closing thsi PR. We still have the bigger problem of #163 ...

bbradbury:
My proposal for #163 would be to eliminate per-model support for factory overriding in favor of reasonable defaults, which could still be overridden in accessors as necessary. I think this is closer to how database adapters commonly behave in rails, though it does depart from how activerecord-postgis-adapter has worked thus far. This PR differs from #176 in that it sets both a Geographic factory and Cartesian factory by default-- so users will get a working RGeo object for any spatial column without any additional configuration, whether it is a geometry or a geography. Also I think possibly Spatial.geographic? works in more cases with the fix on line 66.

bbradbury:
Ah, and one more use case supported here that I forgot about: calcualted fields. Since the factory is tied to the Postgres column type, rather than a known model attribute, things like

```
    sql = %{SELECT
              "fields".*,
              ST_Buffer("fields".bounds, 100.0) AS buffered_bounds} 

Field.find_by_sql(sql)
f.buffered_bounds
```
work correctly. This is crucial for my use case as I am frequently aggregating and intersecting polygons with ST_Union and intersecting polygons with ST_Intersect inside of select statements, and thus not am not able to define a column/factory combination ahead of time.

teeparham:
@bbradbury Thank you for the feedback.

Yes, eliminating per-model support is exactly the plan with #163 (see the new commits in the PR).

I'm not sure how this will work with calculated attributes. We should add tests for that.

PR #171, opened 2015-04-03
Added default_function to SpatialColumn and used in initialize function,...

mikebobrov:
... so now it can be used correctly by other applications that rely on proper PostgreSQLColumn initialization


teeparham:
Спасибо :+1:

PR #176, opened 2015-04-26
Add optional coordinates (z,m) for spherical_factory

vvalgis:
Add optional coordinates (z,m) for spherical_factory when Spatial value initializing.


teeparham:
This improves the default factory. Thanks!

PR #163, opened 2015-04-26
Column specific factory settings are ignored

teeparham:
We just upgraded our application to Rails 4.2 and also to activerecord-postgis-adapter 3.0.0.beta1 and ran into a problem I want to report here.

Before the upgrade the custom factories we set per column worked really well:

```
set_rgeo_factory_for_column(:latlon, RGeo::Geographic.spherical_factory(:srid => 4326))
```

Now the factory setting gets ignored. We always get back "CAPI"-Implementations and not the implementations of the specified factory when reading from the database.

I dug into the code of activerecord-postgis-adapter and built a test case:

```ruby
  def test_loading_with_custom_factory
    klass = SpatialModel
    klass.connection.create_table(:spatial_test, force: true) do |t|
      t.spatial(:latlon, srid: 4326, type: 'point')
    end

    factory = RGeo::Geographic.spherical_factory(srid: 4326)
    klass.class_eval do
      set_rgeo_factory_for_column(:latlon, factory)
    end

    obj = SpatialModel.new
    obj.latlon = 'POINT(1 2)'
    obj.save!

    id = obj.id
    obj2 = SpatialModel.find(id)
    assert_equal factory.class, obj2.latlon.factory.class
  end
```

This fails:

```
  1) Failure:
BasicTest#test_loading_with_custom_factory [test/basic_test.rb:119]:
Expected: RGeo::Geographic::Factory
  Actual: RGeo::Geos::CAPIFactory
```

And that's exactly the problem we have in our application.

It seems like the factory settings are stored in `SpatialColumn` and the parsing is done in `OID::Spatial`. In `OID::Spatial` the factory settings are not used:

```ruby
def parse_wkt(string)
  # factory = factory_settings.get_column_factory(table_name, column, constraints)
  factory = @factory_generator || RGeo::ActiveRecord::RGeoFactorySettings.new
  wkt_parser(factory, string).parse(string)
rescue RGeo::Error::ParseError
  nil
end
```

The line that's commented out is exactly the one we need. But un-commenting it doesn't do anything but break the code. `OID::Spatial` has no access to the `table_name` and `column` and `constraints`.

Do you have any ideas regarding this problem? Do you want to leave it as it is, since it seems difficult to solve with the new PostgreSQL adapter architecture in Rails 4.2, or try to fix it? Any idea what I can try to help?

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,50 +288,29 @@ database has an integer type, ActiveRecord automatically casts the data to a
 Ruby Integer. In the same way, the activerecord-postgis-adapter automatically
 casts spatial data to a corresponding RGeo data type.
 
-However, RGeo offers more "flexibility" in its type system than can be
+RGeo offers more flexibility in its type system than can be
 interpreted solely from analyzing the database column. For example, you can
 configure RGeo objects to exhibit certain behaviors related to their
 serialization, validation, coordinate system, or computation. These settings
-are embodied in the RGeo "factory" associated with the object.
+are embodied in the RGeo factory associated with the object.
 
-Therefore, you can configure the adapter to use a particular factory (i.e. a
-particular combination of settings) for data associated with each column in
-the database. This is done by calling class methods on the ActiveRecord class
-associated with that database table. Specifically, you can call
-`set_rgeo_factory_for_column` to set the factory that ActiveRecord uses for a
-particular column.
-
-You can also provide a "factory generator" function which takes information
-from the database column and returns a suitable factory. Set the factory
-generator by setting the `rgeo_factory_generator` class attribute of your
-ActiveRecord class. The generator should be a callable object that takes a
-hash that could include the following keys:
-
-*   `:srid` -- the SRID of the database column
-*   `:has_z_coordinate` -- true if the database column has a Z coordinate
-*   `:has_m_coordinate` -- true if the database column has a M coordinate
-*   `:geographic` -- true if the database column is geographic instead of
-    geometric
-
+You can configure the adapter to use a particular factory (i.e. a
+particular combination of settings) for data associated with each type in
+the database.
 
 Here are some examples, given the spatial table defined above:
 
 ```ruby
-class MySpatialTable < ActiveRecord::Base
+# on application setup. For example, in an initializer:
 
+RGeo::ActiveRecord::PostGISAdapter::SpatialFactoryStore.instance.tap do |config|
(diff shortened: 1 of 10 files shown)


teeparham:
We still need to deprecate the old `set_rgeo_factory_for_column` and `rgeo_factory_for_column` methods, since they will currently run and do nothing, which is deceptive. I'd rather not include those methods from rgeo-activerecord at all. Maybe that stuff should be removed from rgeo-activerecord altogether.

Thoughts on this PR? cc @bbradbury @mikeycgto @ThomasPr

- [ ] Remove `rgeo_factory_for_column` methods from rgeo-activerecord
- [ ] Update readme with instructions for registering spatial factories for column types

mrnugget:
So, if I understand it correctly, I can set global factories per column type in an initializer/config file, right? I like it! 



> We still need to deprecate the old set_rgeo_factory_for_column and rgeo_factory_for_column methods, since they will currently run and do nothing, which is deceptive. I'd rather not include those methods from rgeo-activerecord at all. Maybe that stuff should be removed from rgeo-activerecord altogether.

I'm all for removing them. Since this would be in the `3.x.x` release anyway, a major release, I don't think there is a problem with introducing breaking changes as long as the change is documented and clear instructions on how to upgrade are provided.

ThomasPr:
I totally agree with @mrnugget. It's a major release, API changes are ok as long as they're documented.

teeparham:
@mrnugget That's correct. Give it a try.

I think I will move the `SpatialFactoryStore` class to `rgeo-activerecord`, as it's a replacement for the `factory_for_column*` methods/class, and could be used by other adapters. I'll take a closer look tonight.

teeparham:
Closing in favor of #180. I moved `SpatialFactoryStore` to `rgeo-activerecord`.

mrnugget:
Just FYI: we upgraded to beta5 and 3.0.0 of activerecord-rgeo without any problems. I threw out the monkey-patches I had and replaced them with two lines of configuration to specify the default factory. Very cool. Great work! Thanks! :)

teeparham:
@mrnugget thanks for testing & updating this issue. Good to hear!

teeparham:
This points to a solution for defining a custom Type on an attribute. The Type could then have its own spatial factory.

https://blog.nvisium.com/2015/06/using-rails-5-attributes-api-today-in.html

PR #194, opened 2015-08-16
Could we start Rails 5 branch?

dzjuck:
I'm trying migrate one project to Rails 5 and as part of it i added support of current Rails 5 to activerecord-postgis-adapter. This pull request contains code are working with Rails 5.

Could we create branch for Rails 5 version and do development there, while Rails 5 will not be released? Also we need Rails 5 branch for rgeo-activerecord, so how it depends on ActiveRecord.

If anybody have question or comments about code - i will glad to answer. Thank you.


teeparham:
There is now a `rails-5` branch. Please do any rails 5 work based on that branch & submit PRs to that branch.

Issue #198, opened 2015-11-18
JRuby support


deivid-rodriguez:
In the README it is stated that JRuby is not supported because jruby's ActiveRecord adapter does not support 4.2 yet. But now it does, I think.

I tried running the test suite against `jruby 9.0.3.0` and I only got two failures.

```
deivid@pantani ~/code/activerecord-postgis-adapter $ chruby jruby
deivid@pantani ~/code/activerecord-postgis-adapter $ bundle exec rake appraisal
uri:classloader:/jruby/kernel/kernel.rb:17: warning: unsupported exec option: close_others
uri:classloader:/jruby/kernel/kernel.rb:17: warning: unsupported exec option: close_others
>> BUNDLE_GEMFILE=/home/deivid/code/activerecord-postgis-adapter/gemfiles/ar42.gemfile bundle exec rake
uri:classloader:/jruby/kernel/kernel.rb:17: warning: unsupported exec option: close_others
Run options: --seed 23018

# Running:

...............................E.......F..................

Fabulous run in 15.514597s, 3.7384 runs/s, 10.6996 assertions/s.

  1) Error:
NestedClassTest#test_nested_model:
NoMethodError: undefined method `result_status' for 0:Fixnum
    /home/deivid/code/activerecord-postgis-adapter/test/nested_class_test.rb:18:in `test_nested_model'


  2) Failure:
DDLTest#test_column_defaults [/home/deivid/code/activerecord-postgis-adapter/test/ddl_test.rb:266]:
Expected: "-1"
  Actual: -1

58 runs, 166 assertions, 1 failures, 1 errors, 0 skips
rake aborted!
Command failed with status (1): [ruby -I"lib:test" -I"/home/deivid/.gem/jruby/2.2.2/gems/rake-10.4.2/lib" "/home/deivid/.gem/jruby/2.2.2/gems/rake-10.4.2/lib/rake/rake_test_loader.rb" "test/**/*_test.rb" ]

Tasks: TOP => default => test
(See full trace by running task with --trace)
```

They seem naively like easy fixes. Maybe someone with more knowledge can have a look?

/cc @teeparham @kares

vjdhama:
Also JRuby team [confirmed](https://github.com/jruby/activerecord-jdbc-adapter/issues/692#issuecomment-158046088) that ` AR-JDBC not supporting 4.2 yet.` is no longer true as it's stated in README.

teeparham:
[...]

PR #199, opened 2015-11-19
Jruby support

deivid-rodriguez:
Hi! This is an attempt to get the tests passing for JRuby. The main issue is 3d007649bbfa99d6cff4faeeea05e8aa8567d7af, since JRuby and MRI behave differently here and I don't know what's the expected behavior and whether this should be considered public API at all. 

* JRuby's behavior is actually tested: https://github.com/jruby/activerecord-jdbc-adapter/blob/v1.3.19/test/db/postgresql/simple_test.rb#L173-L180
* I haven't found any tests or documentation for the opposite behavior in `activerecord` but I think it's intentional. I think it happens here: https://github.com/rails/rails/blob/7f18ea14c893cb5c9f04d4fda9661126758332b5/activerecord/lib/active_record/connection_adapters/postgresql/quoting.rb#L59-L68

Opinions?

EDIT: Just linking to the original issue: #198

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ rvm:
   - 2.0.0
   - 1.9.3
   - jruby-1.7
+  - jruby-9.0.4.0
   - rbx-2
 gemfile:
   - gemfiles/ar42.gemfile
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ rvm:
   - 2.1
   - 2.0.0
   - 1.9.3
-  - jruby-19mode
+  - jruby-1.7
   - rbx-2
 gemfile:
   - gemfiles/ar42.gemfile
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ gemfile:
 matrix:
   allow_failures:
     - rvm: rbx-2
-    - rvm: jruby-19mode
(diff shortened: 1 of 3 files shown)


teeparham:
Thank you! :tada: :clap:

deivid-rodriguez:
Fuzzy matching does not work... `jruby-9.0` is matched against a prerelease version, while just `jruby` matches it against `jruby-1.7`. We need to be explicit.

deivid-rodriguez:
Thanks! :smile: 

I'd like to ask @kares though about this difference in behavior because whether this `klass.columns` trick is supported or not, I've seen it used in some places, so it'd be good to know what to expect.

kares:
hey! honestly do not know if its much of an issue - recall noticing some `column.default` incompatibilities with MRI on 4.2 but I did not care much (we get type-casted results from elsewhere as well) esp. as the same column code supports older AR versions and no one objected so far. so it is as it is until it is :hear_no_evil:

deivid-rodriguez:
Thanks! Sounds reasonable!

teeparham:
This PR is included in the 3.1.0 gem release: https://rubygems.org/gems/activerecord-postgis-adapter

JRuby folks: please give it a try & report any issues.

PR #208, opened 2016-02-05
Ignore PostGIS views when schema dumping

derekprior:
Rails has no support for views natively, but several gems ([Scenic]
included), do add support for them. Using them with PostGIS can cause
the PostGIS views themselves to be dumped.

The PostGIS adapter already ignores its own tables. This change adds the
view names that I'm aware of to the list.

See: https://github.com/thoughtbot/scenic/issues/134

[Scenic]: https://github.com/thoughtbot/scenic

### DIFF
--- a/lib/active_record/connection_adapters/postgis/setup.rb
+++ b/lib/active_record/connection_adapters/postgis/setup.rb
@@ -2,7 +2,15 @@ module ActiveRecord  # :nodoc:
   module ConnectionAdapters  # :nodoc:
     module PostGIS  # :nodoc:
       def self.initial_setup
-        ::ActiveRecord::SchemaDumper.ignore_tables |= %w(geometry_columns spatial_ref_sys layer topology)
+        ::ActiveRecord::SchemaDumper.ignore_tables |= %w(
+          geography_columns
+          geometry_columns
+          layer
+          raster_columns
+          raster_overviews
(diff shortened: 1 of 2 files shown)


lib/active_record/connection_adapters/postgis/setup.rb
```diff
@@ -2,7 +2,16 @@ module ActiveRecord  # :nodoc:
...
     module PostGIS  # :nodoc:
       def self.initial_setup
-        ::ActiveRecord::SchemaDumper.ignore_tables |= %w(geometry_columns spatial_ref_sys layer topology)
+        ::ActiveRecord::SchemaDumper.ignore_tables |= %w(
+          geometry_columns
+          spatial_ref_sys
+          layer
+          topology
+          geography_columns
+          geometry_columns
```
teeparham: `geometry_columns` is in the list twice. Can you de-dupe & alphabetize these? Thanks!

derekprior: Amended

derekprior:
CI Failure looks similar to what I reported here on other projects: https://github.com/travis-ci/travis-ci/issues/5580

Don't think it's related to this change at all.

teeparham:
:+1: Thanks

PR #209, opened 2016-02-08
fixed valid conn params

sespindola:
The ActiveRecord::ConnectionHandling module doesn't define the VALID_CONN_PARAMS anymore,
so I've modified it to obtain the valid parameters hash from PG.


teeparham:
Thanks for this PR. Closing in favor of 44347815ae, which replaces the method based on the upstream AR code.

PR #212, opened 2016-02-14
Add support for activerecord 5.0.0.beta

dzjuck:
It works with activerecord-5.0.0.beta1! Tests are green.

Waiting for response


teeparham:
Thank you very much!

One suggestion: next time, can you break your work in to smaller commits?

:clap:

dzjuck:
Sure, next time i will break process to separate commits

PR #213, opened 2016-02-18
Support ActiveRecord 5.0

teeparham:
Support for activerecord 5.0


danascheider:
Thank you for this, it is saving me a lot of PITA.

PR #220, opened 2016-04-19
Add missing configuration option to README

stevejackson:
This is an option that appears to have been supported since 2.x, but is missing from the documentation.

Because my database uses postgis within the `postgis` schema instead of `public`, this configuration option was necessary when upgrading to 2.x.


teeparham:
Thanks!

PR #223, opened 2016-04-29
adjusted call to new TableDefinition due to rails change

jklimke:
There was a shift active record TableDefinition towards using keyword arguments. (noiticed when trying out with rails 5 beta)

I fixed the incompatibility  by passing the arguments of create_table_definition directly to the initialize method of the PostGIS::TableDefinition


jklimke:
actually this change was already included in the rails 5 pull request

PR #226, opened 2016-05-19
fixing schema dumping for active record 4.2

mikebobrov:
As said in the documentation: "In ActiveRecord 4.2, the Postgresql adapter added support for the native Postgresql point and polygon types, which conflict with this adapter's types of the same names.". Rails SchemaDumper used incorrect type for generated column definitions, e.g.:
t.geography "lonlat", limit: {:srid=>4326, :type=>"point", :has_z=>true, :geographic=>true}

This fix sets correct types st_point and st_polygon. So with that fix Schema Dumper gives:
t.geography "lonlat", limit: {:srid=>4326, :type=>"st_point", :has_z=>true, :geographic=>true}

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ Gemfile.lock
 /gemfiles/*.gemfile.lock
 /travis/*.lock
 /test/database_local.yml
+.idea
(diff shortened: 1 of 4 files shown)


teeparham:
Closing in favor of #252, since I want to exclude the `.gitignore` & version dependency code. With smaller commits, I could cherry-pick the commits.

Issue #175, opened 2015-04-24
Schema.rb + UUID breaks


ndbabb:
I'm having an issue where my schema in schema.rb is altered incorrectly. 

I added the gem to an existing rails 4.2 project running postgresql 9.3.6. I've installed postgis 2.1.2.

Note that I'm also using the uuid-ossp extension so that my rails models can utilize a UUID-based id column for the primary key.

After install the gem (3.0.0.beta2) and updating my database.yml accordingly, I will run the next migration (or just run rake db:schema:dump) and every table in my schema.rb has its "default" option changed from "uuid_generate_v4()" to nil, e.g.

```
#original
create_table "users", id: :uuid, default: "uuid_generate_v4()", force: :cascade do |t| 

#after postgis and rake db:schema:dump
create_table "users", id: :uuid, default: nil, force: :cascade do |t|

```

I backed out the gem and related changes and then tried to manually enable postgis ("CREATE EXTENSION postgis;"), followed by "rake db:schema:dump" and the issue does not occur. 

Thanks in advance for any advice.

desaperados:
I raised the above Rails ticket as I first thought this was an issue with Rails. However after a bit more investigation have isolated the problem to the postgis adapter. We're also using uuid's and getting default: nil incorrectly dumped into schema.rb

michaelkelly322:
I just ran into this issue today, it looks like if you update from `3.0.0.beta2` to the full `3.0.0` release the issue is resolved.

desaperados:
Can confirm that updating to the full release resolves the issue

teeparham:
@michaelkelly322 @desaperados thanks for reporting on this issue.

Closing as fixed in 3.0.0.

PR #230, opened 2016-06-23
Call to super does not match initialize parameters

jmera:
SpatialColumn inherits from PostgreSQLColumn (which inherits from [`Column`](https://github.com/rails/rails/blob/e8c21ff645eefb889acd7799bd12c9757cd8fd8c/activerecord/lib/active_record/connection_adapters/column.rb)) and it's call to super should match it's own initialize signature.

This change is related to rgeo/activerecord-postgis-adapter#175 because under the right circumstances, `default_function` is passed to super with a nil value setting the wrong default_function.

@teeparham I'd like to hear your thoughts on how to test this out...

### DIFF
--- a/lib/active_record/connection_adapters/postgis/schema_statements.rb
+++ b/lib/active_record/connection_adapters/postgis/schema_statements.rb
@@ -31,6 +31,7 @@ module ActiveRecord
                             default,
                             sql_type_metadata,
                             null,
+                            table_name,
                             default_function,
                             collation,
                             cast_type,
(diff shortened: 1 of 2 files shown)


teeparham:
Thank you. This looks good.

Yes, the method signature of `Column` was changed here:

https://github.com/rails/rails/commit/6b6a01c52

I'm open to suggestions on how to test it. I ran the tests yesterday with the latest activerecord 5.0.0.rc2 & it worked, so that's not good.

It would probably go in `test/ddl_test.rb` & pass a `default: "something"` to a column definition. For reference: https://github.com/rails/rails/blob/master/activerecord/test/cases/adapters/postgresql/uuid_test.rb

Issue #236, opened 2016-11-07
Memory leak with postgis adapater & Sidekiq


inkstak:
Hello, I'm investigating on a memory leak. You can found a rails app [here](https://github.com/inkstak/memory_leak_test) to reproduce.

I'm running this [job](https://github.com/inkstak/memory_leak_test/blob/master/app/jobs/mem_test_job.rb) to browse a large table with the Rails method `find_each`.
In this example, I try to call `GC.start` every 10 batches to reduce memory impact and then inspect the GC stats and current memory usage.

When the job starts:
```
GC.start                        After         Before           Last
 :heap_available_slots        324,855        324,855        309,774
 :heap_live_slots             246,979        323,461        164,291
 :heap_free_slots              77,876          1,394        145,483
Memory: 75.9 MB
```

After dozens of occurences:
```
GC.start                        After         Before           Last
 :heap_available_slots     12,670,227     12,670,227     12,049,043
 :heap_live_slots          12,621,392     12,669,837     12,000,422
 :heap_free_slots              48,835            390         48,621
Memory: 815 MB
...
GC.start                        After         Before           Last
 :heap_available_slots     13,292,225     13,292,225     12,670,227
 :heap_live_slots          13,242,362     13,291,861     12,621,392
 :heap_free_slots              49,863            364         48,835
Memory: 851 MB
```

The heap size and the memory usage increase until the final blow up.
The leak happens when using the postgis adapter through Sidekiq:
* performing the job through the rails console with the gem 'activerecord-postgis-adapter' in Gemfile doesn't cause a memory leak
* performing the job through the sidekiq without the gem 'activerecord-postgis-adapter' in Gemfile 
doesn't cause a memory leak
* performing the job through the sidekiq with the gem 'activerecord-postgis-adapter' in Gemfile cause a memory leak

[EDIT: Memory usage reported was in MB, not in KB]

inkstak:
The issue comes from the loading of `ActiveRecord::ConnectionAdapters::PostGIS::PostGISDatabaseTasks` which inherits from `ActiveRecord::Tasks::PostgreSQLDatabaseTasks`

I guess that `PostgreSQLDatabaseTasks` is not multithread-prof, that why ActiveRecord use autoloading (provided by ActiveSupport) to require it only on needed : when calling rails tasks.

teeparham:
Fixed by #237

PR #237, opened 2016-11-07
Autoload PostGISDatabaseTasks to avoid memory leaks in multithread

inkstak:
Issue described in #236
PostGISDatabaseTasks is not multithread-prof. It should be required on needed.

### DIFF
--- a/lib/active_record/connection_adapters/postgis_adapter.rb
+++ b/lib/active_record/connection_adapters/postgis_adapter.rb
@@ -15,7 +15,6 @@ require "active_record/connection_adapters/postgis/arel_tosql"
 require "active_record/connection_adapters/postgis/setup"
 require "active_record/connection_adapters/postgis/oid/spatial"
 require "active_record/connection_adapters/postgis/create_connection"
-require "active_record/connection_adapters/postgis/postgis_database_tasks"
 
 ::ActiveRecord::ConnectionAdapters::PostGIS.initial_setup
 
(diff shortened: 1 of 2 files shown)


teeparham:
Thank you. Can you move the autoload code into `lib/active_record/connection_adapters/postgis_adapter.rb` file & use double quotes? That way, if people require either file, things will work.

inkstak:
Updated

cgthornt:
FYI this appears to have broke my rake tasks in Rails 5.0.0.1 and ruby 2.1.3:

```
$ rake db:drop
Rake tasks not supported by 'postgis' adapter
Couldn't drop database 'my_db'
rake aborted!
ActiveRecord::Tasks::DatabaseNotSupported: Rake tasks not supported by 'postgis' adapter

Tasks: TOP => db:drop:_unsafe
```

teeparham:
@cgthornt thanks for reporting... I'd go back to 4.0.0 until this is sorted out. (I think you mean ruby 2.3.1, btw).

I bet it's because this line is outside the class definition: `::ActiveRecord::Tasks::DatabaseTasks.register_task(/postgis/, PostGISDatabaseTasks)`

inkstak:
Hum, it's complicated. `ActiveRecord::Tasks::DatabaseTasks` is also autoloaded and only loaded when when running the database tasks (create/drop/...)

All these tasks depend on the other task `db:load_config`.
It's not ideal, but enhance this task may be a solution. What do you think ?

teeparham:
I reverted this commit in 4bbe4fe and released version 4.0.2 without the autoloading tasks change.

1) It certainly breaks the tasks, so I removed the autoload change until we have something that we know works.
2) Is the memory leak a real problem?

inkstak:
I investigated deeper about this issue. The problem seems to be elsewhere : https://github.com/rails/rails/issues/27002
The root cause of the leak I described is the ActiveRecord query cache which is, I suppose, enabled in console by a cascade effect when PostGISDatabaseTasks is loaded.

I guess you can cancel this PR.
Sorry for the troubles due to my early report.

teeparham:
@inkstak thank you for the update.

PR #239, opened 2016-11-15
Avoid spaces in the beginning of schema names

caron10:
When schema contains multiple values, for example "schema1, schema2", names will contain leading space (" schema2") after splitting.
This patch fixes this issue.


keithdoggett:
Closing since this is old and does not appear to be an issue on newer versions.

Issue #248, opened 2017-03-24
Rails 5.1 problems


MtnBiker:
With no version specified `bundle update` resulted in `Installing activerecord-postgis-adapter 3.1.2 (was 4.0.2)`, but then `rails server` error 

    ~/.gem/ruby/2.4.0/gems/bundler-1.14.6/lib/bundler/runtime.rb:94:in `rescue in block (2 levels) in require': There was an error while trying to load the gem 'activerecord-postgis-adapter'. (Bundler::GemRequireError)
    Gem Load Error is: uninitialized constant ActiveRecord::ConnectionAdapters::PostgreSQL::ColumnDefinition
    Did you mean?  ActiveRecord::ConnectionAdapters::ColumnDefinition
                   ActiveRecord::ConnectionAdapters::AddColumnDefinition
    Backtrace for gem load error is:
    ~/.gem/ruby/2.4.0/gems/activerecord-postgis-adapter-3.1.2/lib/active_record/connection_adapters/postgis/spatial_table_definition.rb:46:in `<module:PostGIS>'
...
Put version 4.0.2 in Gemfile and got this `bundle update` error

    Bundler could not find compatible versions for gem "activerecord":
      In Gemfile:
        activerecord-postgis-adapter (= 4.0.2) was resolved to 4.0.2, which depends on
          activerecord (~> 5.0.0)
    
        rails (~> 5.1.0.rc1) was resolved to 5.1.0.rc1, which depends on
          activerecord (= 5.1.0.rc1)

MtnBiker:
Why is it closed? Rails people say it doesn't support 5.1 based on https://github.com/rgeo/activerecord-postgis-adapter/blob/master/activerecord-postgis-adapter.gemspec#L22.

If I understand it that file needs to support activerecord 5.1.

scottfromsf:
I agree with @MtnBiker, this seems like an issue that needs to be addressed here, not on the Rails project.

MtnBiker:
@scotfromsf The Rails project agrees. https://github.com/rails/rails/issues/28551. Thank you for commenting.

aomega08:
Definitely an issue here, I started digging into this, but honestly I'm not so well knowledgeable about the AR internals and I got lost in short. BTW the few edits I made on my local clone would not be backward compatible.

But having this before Rails 5.1 is out would be great.

kofronpi:
[...]

PR #246, opened 2017-02-28
ActiveRecord 5.1

glaszig:
just started a new project on rails 5.1.0.beta1 and i need rgeo to work.
first time for me to mock around with AR in any meaningful way -- sorry if i'm screwing things.
tests are green on my side with postgresql 9.5 on ubuntu 16.04.

PR #251, opened 2017-04-26
Allow rgeo-activerecord 5.x

aomega08:
First step to support ActiveRecord 5.1. ( #248 )

The rgeo-activerecord gem seems fine with AR 5.1.


aomega08:
Just noticed this is already covered by #246 , so feel free to close this. I was going on a more step-by-step approach.

PR #261, opened 2017-09-01
fixed Postgres + PostGIS on Travis CI

kyohei-shimada:
This PR is a solution same as https://github.com/boundlessgeo/exchange/pull/111


teeparham:
Thank you!

PR #271, opened 2017-12-01
Add support for Rails 5.2

claudiob:
The signature of the private method `SchemaStatements#initialize_type_map`
was changed in https://github.com/rails/rails/pull/29869.

This change reflects the new signature which is supported both by Rails 5.1
and by Rails 5.2.


teeparham:
Thank you!

I will release a new gem version when Rails 5.2.0 is released.

For now, to use with Rails 5.2.0.beta:

```
gem "activerecord-postgis-adapter", git: "https://github.com/rgeo/activerecord-postgis-adapter.git"
```

alhafoudh:
Can we please get rails 5.2.0 support? Thanks

teeparham:
@alhafoudh Rails 5.2 has been supported by this gem for 5 months, starting with version 5.0.1.

oshanz:
by looking at commit log it is activerecord-postgis-adapter version 5.1 that support rails 5.2, not the 5.0.1

image

Kiza:
This will affect rails 5.0.2 as well if `schema_cache.dump` is used.

Issue #280, opened 2018-02-27
Bug fixed: when using geometry with has_z


eliasbaixas:
Hi,

in file 
activerecord-postgis-adapter-5.0.3/lib/active_record/connection_adapters/postgis/oid/spatial.rb

line 30, in parse_sql_type where it says:

if params.size > 1

it should be if params.size >= 1

because for example, I have a column without srid (bc I'm using cartesian space) , so the sql_type becomes:

geometry(PointZ) and since params.size == 1, the PointZ doesn't get parsed bc if params.size == 1 (not >1) and then has_z is left "false" and id doesnt parse PointZ(a b c).

fixed just by replacing >1 for >= 1

teeparham:
Thanks for looking into this. Can you submit a PR with a test?

PR #281, opened 2018-03-01
Fixes bug #280

eliasbaixas:
Hello, here's my pull request fixing #280 

hope everyting is fine (it's my first pull request :P )


teeparham:
I think we can get rid of `if params >= 1` on line 30 and the else block altogether. However, this works for now. Thanks!

PR #287, opened 2018-05-30
Support for active record 5.2

RicardoTrindade:
Unlocks active record version in gemspec to support version 5.2


teeparham:
Thanks, but gem version 5.0.1 and later already work with ActiveRecord 5.2. This change is not necessary. See #271

`~> 5.1` means any version `5.x.*` greater than or equal to `5.1`

PR #292, opened 2018-10-19
Ensure connection before `PostGISAdapter#initialize`

kamipo:
Since https://github.com/rails/rails/pull/34232, early checking server version in superclass so no longer accept nil connection.


teeparham:
Thank you. 👍 

The master branch is currently targeting ActiveRecord 5.1 & 5.2, so I need to create new new branch targeting 6.0, & update the upstream method to match 6.0 & update the version dependencies.

teeparham:
Fixed in #303

PR #315, opened 2020-07-08
Adapt gem to ActiveRecord 6.0.

phlegx:
- Clean up class `PostGISAdapter` for ActiveRecord 6.


keithdoggett:
This looks good, thanks @phlegx !

Issue #183, opened 2015-05-28
Arel 6 - geographic changes to geometry in version 3


jaybloke:
Not sure if this is a bug… or I'm simply not using v3 of the adapter correctly ( :-1: )

Before I was using Rails 4.2.1 with version 2.2.2 activerecord-postgis-adapter

```
latlng = RGeo::Geographic.spherical_factory(srid: 4326).point(121.0199964, 14.5380189)
Arel::Nodes::NamedFunction.new('ST_Distance', [latlng, latlng]).to_sql
```

This correctly gives me this SQL:

```
ST_Distance('0101000020E61000006A02FB9E47415E408242983677132D40', ‘0101000020E61000006A02FB9E47415E408242983677132D40')
```

After upgrading to Rails 4.2.1 w/activerecord-postgis-adapter 3 I tried the following:

```
latlng = RGeo::Geographic.spherical_factory(srid: 4326).point(121.0199964, 14.5380189)
Arel::Nodes::NamedFunction.new('ST_Distance', [latlng, latlng]).to_sql
```

But it gives me: RuntimeError: unsupported: RGeo::Geographic::SphericalPointImpl

However, 

```
Arel::Nodes::NamedFunction.new('ST_Distance', [latlng.as_text, latlng.as_text]).to_sql
```

Gives me this SQL:

```
ST_Distance(ST_GeomFromEWKT('POINT (121.0199964 14.5380189)'), ST_GeomFromEWKT('POINT (121.0199964 14.5380189)’))
```

Seems like all geography factories are being converted to geometry in Arel. The workaround is to cast to geography like so:

```
ST_Distance(Geography(ST_GeomFromEWKT('POINT (121.0280889 14.5491603)')), Geography(ST_GeomFromEWKT('POINT (121.0280889 14.5491603)')))
```

This works, but does seem somewhat longwinded. Any pointers or suggestions?

Thx

jaybloke:
It looks like this goes deeper. I set up a basic app and followed the instructions on the readme page and tried this from the console:

~~~
 Arel::Nodes::NamedFunction.new('ANYTHING', ['param_1', 'param_2']).to_sql
~~~

And surprisingly got this: 

~~~
ANYTHING(ST_GeomFromEWKT('param_1'), ST_GeomFromEWKT('param_2'))
~~~

zaidakram:
Yup, facing the same issue here...

```
[...]

PR #318, opened 2020-08-28
Update Arel Module

### LINKED ISSUE #207: Suppresed arel exception, wrong sql
Hi!

I tried to access json fields with `Arel::Nodes::InfixOperation.new('->>', arel_table[:attr], 'field')`. With `adapter: postgis` it doesn't raise exception but produces wrong sql: `"table"."attr" ->> ST_GeomFromEWKT('field') =`. Accessing json arrays with same operation (int instead of string) worked fine.

With `adapter: postgresql` it raises exception from arel `RuntimeError: unsupported: String`. I had an error in my code, and then fixed it by replacing string with `Arel::Nodes.build_quoted('field')` (sql became `"table"."attr" ->> 'field' =`).

I think it'll be better to bypass this exception (if it's possible of course).

keithdoggett:
Fixes #207 , #183 

Removes the `visit_String` method that was causing all strings passed directly into an `Arel` method to be wrapped with `ST_GeomFromEWKT`. Since `visit_String` was removed from `Arel` core, the expected behavior is that passing a string directly into a method will raise an error.

There is probably some work that can be done on `rgeo-activerecord` in regards to its `Arel` interface, but this at least gets the adapter up to date with current `Arel` behavior.

### DIFF
--- a/lib/active_record/connection_adapters/postgis/arel_tosql.rb
+++ b/lib/active_record/connection_adapters/postgis/arel_tosql.rb
@@ -20,10 +20,6 @@ module Arel  # :nodoc:
         FUNC_MAP[standard_name.downcase] || standard_name
       end
 
-      def visit_String(node, collector)
-        collector << "#{st_func('ST_WKTToSQL')}(#{quote(node)})"
-      end
-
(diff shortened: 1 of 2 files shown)


keithdoggett:
Closing in favor of #325

PR #324, opened 2020-12-12
Add Active Record 6.1 compatibility

tagliala:
- ActiveRecord Autoload Change: rails/rails@b744372
- Use DatabaseConfig Instead of Hash for Tasks: rails/rails#37291
- Move ActiveRecord::Tasks::DatabaseAlreadyExists to
  ActiveRecord::DatabaseAlreadyExists: rails/rails@69700c9
- Wrap native adapters connection errors: rails/rails#40110

Close: #322, #323

Issue #322, opened 2020-12-10
NameError: undefined local variable or method `configuration' for PostGISDatabaseTasks


tagliala:
Apologies if it is a duplicate, I've searched for "name error", `configuration`, and `hash_configuration`. I've also checked latest issues, PRs and tested master branch

When running specs against Rails 6.1, I'm getting:

```
NameError: undefined local variable or method `configuration' for #<ActiveRecord::ConnectionAdapters::PostGIS::PostGISDatabaseTasks:0x00007fadf0e921e0>
Did you mean?  configuration_hash
```

I think it may depend by rails/rails#37291

javierjulio:
Same here. I was trying to update an app to Rails 6.1 and when running the test suite this error occurs.

tagliala:
Hi @javierjulio I'm using `ar61` branch which contains AR61 support and specs in my application are all green

We just need to wait for 7.0.0 to be released to go back to the stable gem

javierjulio:
@tagliala good to know, thank you!

PR #325, opened 2020-12-18
ActiveRecord 6.1 (v7.0.0)

### LINKED ISSUE #183: Arel 6 - geographic changes to geometry in version 3

### LINKED ISSUE #216: Support getting registered factories by types documented in rgeo-activerecord
In upgrading a client's system from using a much older version of this adapter (3.0.0 beta 2), I had to switch from using `set_rgeo_factory_for_column` to registering a factory. That part was fine, but it took a little bit of debugging to realize why the following code wasn't working:

```rb
config.register(
  RGeo::Geographic.spherical_factory(srid: 4326, uses_lenient_assertions: true),
  sql_type: 'geography')
```

or alternatively, I tried:

```rb
config.register(
  RGeo::Geographic.spherical_factory(srid: 4326, uses_lenient_assertions: true),
  geo_type: 'multi_polygon')
```

It turns out that these have to be the strings as PostGIS knows them:

```rb
geo_type: 'MultiPolygon',
sql_type: 'geography(MultiPolygon,4326)', # in my particular case, although this could vary dramatically depending on your exact configuration for the column in PostGIS
```

I assume the docs in rgeo-activerecord use different strings with the intent of having an API that is compatible between both PostGIS and Spatialite (which I haven't used, so I don't know how it names these things), but I could be wrong.

In any case, it would be helpful if either this adapter used the strings documented in rgeo-activerecord (e.g. 'multi_polygon' instead of 'MultiPolygon') or the README here clearly called out that the particular values will be different when using PostGIS.

It would also be pretty nice if there was some system of hierarchy/inheritance here. e.g:

- Having a specified `geo_type: 'geometry'` match both `geometry` and its subtypes, like `multi_polygon/MultiPolygon`
- Having a specified `sql_type: 'geography'` match any of `geography`, `geography([shape_type])`, `geography([shape_type], [srid])`

…though I know that would require some upstream changes in `RGeo::ActiveRecord::SpatialFactoryStore#factory()`

keithdoggett:
Add ActiveRecord 6.1 compatability along with other breaking changes.

- [x] ActiveRecord 6.1 Compatability (#324)
- [x] Change Spatial Factory Store attribute parsing (fixes #216)
- [x] Simplify Arel Module (fixes #183 #207)
- [x] Update README
- [x] Update History
- [x] Bump version


keithdoggett:
The `SpatialFactoryStore` parsing and Arel simplification are dependent on `rgeo-activerecord` version 7.0.0 which should be released shortly.

keithdoggett:
I've updated the `Arel::Visitors::PostGIS` class to rely entirely on the implementation in rgeo-activerecord (via `RGeo::ActiveRecord::SpatialToSql`). Unfortunately, there is one minor issue that may not end up being a big deal, but I have some options for it regardless.

I switched from `ST_GeomFromEWKT` to `ST_GeomFromText`, and using the latter allows us to handle RGeo features since it has a parameter for WKT and SRID and it is implemented in multiple spatial database systems.

The drawback is that passing EWKT into `ST_GeomFromText` will raise a warning in PostGIS: `WARNING:  OGC WKT expected, EWKT provided - use GeomFromEWKT() for this`. The query will still execute properly, but this could be annoying. One solution is to encourage people to use the `RGeo::WKRep::WKTParser` with `supports_ewkt` enabled and parse all of their EWKT strings before querying them in an `Arel.spatial` node.

Anyways here are the solutions I think could work:

1. Do what I described in the last paragraph and leave it as is. Encourage people to parse EWKT strings before queries (this is what happens when we save features described by WKT).
2. Rewrite the `RGeo::ActiveRecord::SpatialToSql` `visit_in_spatial_context` function to use `ST_GeomFromText` for features and `ST_GeomFromEWKT` for strings.
3. Override `visit_in_spatial_context` in the postgis-adapter.
4. Use a combination of `prepend` and another module to just rewrite the string case of `visit_in_spatial_context`.

I'd prefer 1 or 2. I think 1 is fine since it's not a breaking change but will raise warnings and in the PostGIS [docs](https://postgis.net/docs/manual-3.0/using_postgis_dbmanagement.html#EWKB_EWKT) they note that it shouldn't be relied on. Let me know your thoughts @BuonOmo

BuonOmo:
@keithdoggett 2. doesn't look to hard to implement, and I think having a warning showing often in logs + the fact that the Postgis doc explicitely say we should not rely on that feature are two great reasons to not stay as-is.

keithdoggett:
@BuonOmo I used `prepend` to modify the `visit_in_spatial_context` function with a deprecation warning, but I also figured I could use the WKTParser in there directly to parse the EWKT string to a feature.

Here's the code:

```rb
#lib/activerecord/connection_adapters/postgis/arel_tosql.rb
module RGeo
  module ActiveRecord
    module SpatialToSqlExtension
      def visit_in_spatial_context(node, collector)
        if node.is_a?(String) && node =~ /^SRID=[\d+]{0,};/
          msg = "EWKT Strings are no longer recommended for Arel.spatial queries, please use the RGeo::WKRep::WKTParser class to convert this to an RGeo Feature first."
          ActiveSupport::Deprecation.warn(msg)
          parser = RGeo::WKRep::WKTParser.new(nil, support_ewkt: true)
          node = parser.parse(node)
        end
        super(node, collector)
      end
    end
  end
end
RGeo::ActiveRecord::SpatialToSql.prepend RGeo::ActiveRecord::SpatialToSqlExtension
```
It's possible to leave the parser in there and remove the deprecation warning, but I'm not sure if this would be too much going on behind the scenes and could be confusing if there was an error in the parser. Let me know what you think.

keithdoggett:
This is how we could rely on the existing `RGeo::ActiveRecord::SpatialToSql` module, but add PostGIS specific handlers for EWKT nodes.

```rb
module RGeo
  module ActiveRecord
    ##
    # Extend rgeo-activerecord visitors to use PostGIS specific functionality
    module SpatialToPostGISSql
      def visit_in_spatial_context(node, collector)
        # Use ST_GeomFromEWKT for EWKT geometries
        if node.is_a?(String) && node =~ /SRID=[\d+]{0,};/
          collector << "#{st_func('ST_GeomFromEWKT')}(#{quote(node)})"
        else
          super(node, collector)
        end
      end
    end
  end
end
RGeo::ActiveRecord::SpatialToSql.prepend RGeo::ActiveRecord::SpatialToPostGISSql
```

I'm going to go with this since it uses the PostGIS functions but doesn't entirely override the original `SpatialToSql` module. If you think I should use the first version if this is too confusing let me know, but for now I'm going to start working on the docs.

BuonOmo:
@keithdoggett that is pretty clean, and I think this is a great benefit. The only downside I can foresee is that in the long run this postgis function has a risk of being deprecated. However this is a really long shot. IMHO, we can assume that we are safe here.

Nice work !

README.md
```diff
@@ -304,6 +314,8 @@ PostGIS polygon type must be referenced as `st_polygon`.
 
 ### Configuring ActiveRecord
 
+_NOTE: In version 7.0.0 changes were made so that the attributes in the second argument of `register` will be camel_cased, which may break existing configurations. See [this commit](https://github.com/rgeo/activerecord-postgis-adapter/pull/325/commits/075d8db036c29b88c0e13234affef260d262e76f) for more details._
```
BuonOmo, line 317: I do not find this statment very clear:

- there is no `camelCase` but rather only `snake_case`
- I'm no sure the commit really helps out, maybe a short example would be better ?

keithdoggett: Good catch! I've updated and added an example.

keithdoggett:
It would be bad if they changed how the function worked in the future or deprecated it, but I assume it would have to be in a major version change so we'd likely have ample warning.

tgroshon:
Hiya!  I hit the underlying problem from issue #322 and swapped my gemfile to install from this branch tag and everything worked!  Thanks so much.

Any guess on a timeline for when this will get published?

keithdoggett:
@tgroshon this should be released early next week, thanks for the patience.

tgroshon:
@keithdoggett Looking forward to it. Thanks for all your work!

gemfiles/.bundle/config
```diff
@@ -0,0 +1,2 @@
+---
+BUNDLE_RETRY: "1"
```
BuonOmo, line 2: I think this should not be part of the project, should it be needed for tests, it could be added as a command line argument

BuonOmo (approved):
One configuration details, but nothing so important. LGTM!

gemfiles/.bundle/config
keithdoggett: Good point! Removed it.

keithdoggett:
Thanks @BuonOmo 

This has just been released in version 7.0.0 on RubyGems

PR #330, opened 2021-01-16
Fix JRuby for ActiveRecord 6.1

keithdoggett:
- Modified the `postgis_connection` method to handle the frozen configuration hash.
- Prepended the `ArJdbc::PostgreSQL` adapter module with an `ArJdbc::PostGIS` module that specifies the `ADAPTER_NAME` as PostGIS.
- Changed `BasicTest#test_multi_polygon_column` to accept any `multi_polygon` not just a `CAPIMultiPolygon` since JRuby uses the GEOS FFI.

### DIFF
--- a/lib/active_record/connection_adapters/postgis_adapter.rb
+++ b/lib/active_record/connection_adapters/postgis_adapter.rb
@@ -90,3 +90,21 @@ module ActiveRecord
     end
   end
 end
+
+# if using JRUBY, create ArJdbc::PostGIS module
+# and prepend it to the PostgreSQL adapter since
+# it is the default adapter_spec.
+# see: https://github.com/jruby/activerecord-jdbc-adapter/blob/master/lib/arjdbc/postgresql/adapter.rb#27
+if RUBY_ENGINE == "jruby"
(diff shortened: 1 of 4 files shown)


lib/active_record/connection_adapters/postgis_adapter.rb
```diff
@@ -90,3 +90,21 @@ def quote(value)
...
 end
+
+# if using JRUBY, create ArJdbc::PostGIS module
+# and prepend it to the PostgreSQL adapter since
+# it is the default adapter_spec.
+# see: https://github.com/jruby/activerecord-jdbc-adapter/blob/master/lib/arjdbc/postgresql/adapter.rb#27
+if RUBY_ENGINE == "jruby"
+  module ArJdbc
+    module PostGIS
+      ADAPTER_NAME = 'PostGIS'.freeze
```
BuonOmo, line 101: you should use the top comment `# frozen_string_literal: true` rather than having to freeze every string :)

BuonOmo (approved):
One comment, nothing blocking !

PR #332, opened 2021-01-17
Fix JRuby for ActiveRecord 5.x

keithdoggett:
Similar to #330. Had to fix a few tests and mainly change the Gemfile to not be pinned to a specific version for the `activerecord-jdbcpostgresql-adapter` gem.


gemfiles/.bundle/config
```diff
@@ -0,0 +1,2 @@
+---
+BUNDLE_RETRY: "1"
```
BuonOmo, line 2: I doubt we need that, do we ?

BuonOmo (approved):
LGTM, I have a single question but nothing too important

gemfiles/.bundle/config
keithdoggett: Done

Issue #284, opened 2018-03-24
Unable to create an attribute for an array


volkanunsal:
I'm unable to use the `array` option for attributes when using postgis adapter. I already created [an issue on Rails repo](https://github.com/rails/rails/issues/32324) about this, but I wonder if the solution might be found here.

Here is [the code](https://gist.github.com/volkanunsal/885983978ad2d25684a1e850f22ab188) to reproduce the issue.

rockwellct:
I ran in to this recently as well, but based on the conversation in  https://github.com/rgeo/activerecord-postgis-adapter/issues/267, it seems you can solve it by adding an initializer that registers the types and modifiers

```
module ActiveRecord
  module ConnectionAdapters
    ActiveRecord::Type.register(
      :array,
      PostgreSQL::OID::Array,
      adapter: :postgis
    )

    ActiveRecord::Type.add_modifier(
      { array: true },
      PostgreSQL::OID::Array,
      adapter: :postgis
    )
  end
end
```

dalton-cole:
Running into this on a new app after switching to the postgis adapter.

keithdoggett:
With #334, the adapter will now check for types registered under the `:postgis` or `:postgresql` adapters.

Issue #210, opened 2016-02-08
Joined select returns CAPIPointImpl instead of SphericalPointImpl


lessless:
Hello,

here are two queries, both of which are intended to select coordinates, but one returns `SphericalPointImpl` and other `CAPIPointImpl`.

```
rb(main):018:0> Address.select('coordinates::geometry').first.coordinates
  Address Load (1.3ms)  SELECT  coordinates::geometry FROM "addresses"  ORDER BY "addresses"."id" ASC LIMIT 1
=> #<RGeo::Geographic::SphericalPointImpl:0x3fe3d3af1d64 "POINT (106.0 10.0)">

rb(main):017:0> Realty.joins(:address).select('realties.id, addresses.coordinates::geometry').first.coordinates
  Realty Load (2.4ms)  SELECT  realties.id, addresses.coordinates::geometry FROM "realties" INNER JOIN "addresses" ON "addresses"."addressable_id" = "realties"."id" AND "addresses"."addressable_type" = $1  ORDER BY "realties"."id" ASC LIMIT 1  [["addressable_type", "Realty"]]
=> #<RGeo::Geos::CAPIPointImpl:0x3fe3d486985c "POINT (106.0 10.0)">
```

Might it be a bug?

keithdoggett:
This has been partially fixed by #334. In order to properly cast a geometry from a join, an `attribute` must be registered on the target table. In this example, you would have to do this.

```ruby
class Realty < ActiveRecord::Base
  # ...
  attribute :coordinates, :st_point, srid: 4326, geographic: true
  # ...
end
```

and then the `coordinates` field from a join will be typecast to a geographic point.

keithdoggett:
Closing in favor of #336 since that is more explicit about the issue.

PR #334, opened 2021-02-11
Support Attributes

### LINKED ISSUE #267: Unknown type Jsonb when using Attributes API.
Hi. 

I'm having an issue when declaring a jsonb attribute using the attributes api and the postgis adapter. This works fine when using the `postgresql` adapter. 

If I have a model defined like this

```
class AttributeBug < ApplicationRecord
  attribute :foobar, :jsonb
end
```

I get the following error in rails console 

```
irb(main):003:0> AttributeBug.new
ArgumentError: Unknown type :jsonb
	from (irb):3
irb(main):004:0> AttributeBug.new
=> #<AttributeBug id: nil, created_at: nil, updated_at: nil>
```

Note the 2nd call doesn't fail, although the model does not have the `foobar` attribute defined on it. 

I've created a sample app here to recreate the problem. I tried to dig through this and the `pg` gem to see the difference in how the `ActiveRecord::Type`s are registered, but couldn't figure it out. I'll continue to search on my own but wanted to open an issue on it for now. 

https://github.com/jhirn/postgis_jsonb_attribute_bug/blob/master/README.md

keithdoggett:
This adds support for the standard Postgresql attributes and RGeo features as attributes.

Changes:

- Set the adapter name as `postgresql` in the `type` module that is used for determining which types ActiveRecord should use. This fixes #267 and #284. Now users won't have to redefine Postgres attributes if they want them available in their app, but the downside is that if custom types were registered under `postgis` they will have to be changed to `postgresql`.
- Change the `OID::Spatial` module to accept keywords as arguments (`geo_type`, `srid`, `has_z`, `has_m`, `geographic`) instead of `oid` and `sql_type`. This is a necessary step to allow RGeo features as attributes because attributes access the type directly and do not pass a `sql_type` in.
- Modify `initialize_type_map` to parse `sql_type` then define the OID for each column.
- Register all RGeo column types as attributes using `OID::Spatial`.

With these changes, we can define attributes like this:

```ruby
class SpatialClass < ActiveRecord::Base
  attribute :geo_point, :st_point, srid: 4326, geographic: true
end
```

and it doesn't need an associated column in the database. 

Importantly, this allows for the proper joining of spatial data (#210), although it is not seamless yet. In order to properly format RGeo features from a join, the spatial column needs to be re-defined as an attribute on the target table. For example:

```ruby
class SpatialFoo < ActiveRecord::Base
 belongs_to :foo

 # has column geo_point (:st_point, srid: 4326, geographic: true)
end

class Foo < ActiveRecord::Base
 has_one :spatial_foo
 
 # re-define geo_point here so join works
 attribute :geo_point, :st_point, srid: 4326, geographic: true
end

# perform a query where geo_point is joined to foo
foo = Foo.joins(:spatial_foo).select("foos.id, spatial_foos.geo_point").first
p foo.geo_point.class
# => RGeo::Geographic::SphericalPointImpl
p foo.geo_point.srid
# => 4326
```

TODO:

- [x] Add support for attributes
- [x] Update README with examples
- [x] Update History


lib/active_record/connection_adapters/postgis/oid/spatial.rb
```diff
@@ -11,9 +11,12 @@ class Spatial < Type::Value
           #   "geography(Point,4326)"
           #   "geometry(Polygon,4326) NOT NULL"
           #   "geometry(Geography,4326)"
-          def initialize(oid, sql_type)
-            @sql_type = sql_type
-            @geo_type, @srid, @has_z, @has_m = self.class.parse_sql_type(sql_type)
+          def initialize(geo_type: 'geometry', srid: 0, has_z: false, has_m: false, geographic: false)
```
BuonOmo, line 14: The documentation should be updated as well

lib/active_record/connection_adapters/postgis/oid/spatial.rb
```diff
@@ -43,7 +46,9 @@ def self.parse_sql_type(sql_type)
               # otherType(a,b)
               geo_type = sql_type
             end
-            [geo_type, srid, has_z, has_m]
+            geographic = !(sql_type =~ /geography/).nil?
```
BuonOmo, line 49:
```suggestion
            geographic = sql_type.match? /geography/
```

lib/active_record/connection_adapters/postgis/schema_statements.rb
```diff
@@ -96,8 +96,9 @@ def initialize_type_map(map = type_map)
             st_point
             st_polygon
           ).each do |geo_type|
-            map.register_type(geo_type) do |oid, _, sql_type|
-              OID::Spatial.new(oid, sql_type)
+            map.register_type(geo_type) do |_, _, sql_type|
+              geo_type, srid, has_z, has_m, geographic = OID::Spatial.parse_sql_type(sql_type)
+              OID::Spatial.new(geo_type: geo_type, srid: srid, has_z: has_z, has_m: has_m, geographic: geographic)
```
BuonOmo, line 101: Since I don't know much about this adapter the question may sound stupid. Is `OID::Spatial` a private API only meant to be used by the adapter ? If not this is a major version we'll need to bump

lib/active_record/connection_adapters/postgis/type.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+
+      private
+
+      def current_adapter_name
+        :postgresql
+      end
+    end
+  end
+end
```
BuonOmo, line 29: I really do not grab that one : why do we need to monkey patch `ActiveRecord::Type` rather than just what you did in the next file, registering with correct type `postgresql` ? It feels like this should be under `ConnectionAdapters::PostGIS`

test/attributes_test.rb
```diff
@@ -0,0 +1,100 @@
...
+    assert_equal data.path, data.polygon.exterior_ring
+
+    assert_equal 4326, data.geo_path.srid
+    assert_equal RGeo::Geographic::Factory, data.geo_path.factory.class
+  end
+
+  def test_joined_spatial_attribute
+    # TODO: The attributes that will be joined have to be defined on the
+    # model we make the query with. Ideally, it would "just work" but
+    # at least this workaround makes joining functional.
```
BuonOmo, line 71: Did you plan anything for that ? 🙂

BuonOmo (commented):
Still some todos left on README, and history to update. I've also made some comments, mostly questions to make sure I understood the code correctly.

lib/active_record/connection_adapters/postgis/schema_statements.rb
keithdoggett: It should be a private API only used by the adapter. It defines how we parse data from a column with a SQL definition like `geometry(POINT, 4326)`. Now we're adding the ability to define those same properties with the attributes API. Typically attributes and columns will use the same `OID` type class to keep the type casting consistent for both. 

Here's the attributes registered in the postgres adapter (https://github.com/rails/rails/blob/ee7cf8cf7569ef87079c48ee81c867eae5e24ed4/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L948-L967)

Here's the database types registered in the postgres adapter (https://github.com/rails/rails/blob/ee7cf8cf7569ef87079c48ee81c867eae5e24ed4/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L517-L585).

Though if we didn't want to worry about breaking changes, I could essentially copy the `OID::Spatial` class to an `OID::SpatialAttribute` class and have that handle the attributes only.

lib/active_record/connection_adapters/postgis/type.rb
keithdoggett: The reason we can't just register the types under `postgresql` is that the `ActiveRecord::Type` module returns `:postgis` for `adapter_name_from`. This means that we miss out on all of the `postgresql` specific types (https://github.com/rails/rails/blob/ee7cf8cf7569ef87079c48ee81c867eae5e24ed4/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L948-L967) even though they're compatible with PostGIS.

One option could just be to re-register those above types and leave our name as `:postgis` in the type module, but then we would have to make sure we're up to date with the types every time there is an update to ActiveRecord (not that big of a deal, but still an additional thing to check).

Does that answer your questions?

test/attributes_test.rb
keithdoggett: I'll open up an issue once this is approved :sweat_smile:

lib/active_record/connection_adapters/postgis/oid/spatial.rb
keithdoggett: Since I'm expecting `geographic` to be a boolean I originally had `!!sql_type.match? /geography/` but rubocop suggested my edit. It will probably work with `nil` as well though.

lib/active_record/connection_adapters/postgis/oid/spatial.rb
keithdoggett: Yes, good point I'll do that!

lib/active_record/connection_adapters/postgis/oid/spatial.rb
BuonOmo: You do not need the `!!` since `String#match` returns only true or false, and if `sql_type` can be null then you can use `/geography/.match? sql_type`

lib/active_record/connection_adapters/postgis/schema_statements.rb
BuonOmo: If this is all internal there is nothing to worry about, at most we can make it a minor..

Thanks for the explanation, it is much more clear !

lib/active_record/connection_adapters/postgis/type.rb
BuonOmo: Partially : in the example you gave me we can see that postgresql is not guessed from any method. And what is bothering me is that we monkey patch some active_record common stuff, not under any namespace. Hence I am worried that we may have conflict with other gems.. But there may still be something I don't fully understand...

lib/active_record/connection_adapters/postgis/oid/spatial.rb
keithdoggett: Ahh yeah you're right. I was using `match` not `match?` so that's where the confusion came from. Will just use `match?`.

lib/active_record/connection_adapters/postgis/type.rb
keithdoggett: Yeah, that makes sense. Here's the original `adapter_from_name` method (https://github.com/rails/rails/blob/beed933b9bb9f317697a95a36da067a67797e0d8/activerecord/lib/active_record/type.rb#L49-L51), and here are all the places it is used (https://github.com/rails/rails/search?q=adapter_name_from), so we can see it's pretty limited in its usage. 

That being said, you're right it's probably not ideal to be modifying the ActiveRecord internals since other gems could be relying on it being consistent with the adapter name. I think the best solution is to just copy the postgres specific types and register them under `:postgis` in this adapter.

BuonOmo (commented):
That looks great, I added one last review, tell me what you think about it. Otherwise we'll be ready to merge soon 😄

lib/active_record/connection_adapters/postgis/type.rb
```diff
@@ -0,0 +1,20 @@
...
+  module ConnectionAdapters
+    module PostGIS
+      module Type
+        # Look for :postgis types first, then check for :postgresql
+        # types to simulate a kind of Type inheritance.
+        def lookup(*args, adapter: current_adapter_name, **kwargs)
+          registry.lookup(*args, adapter: adapter, **kwargs)
+        rescue ArgumentError
+          super(*args, adapter: :postgresql, **kwargs)
+        end
```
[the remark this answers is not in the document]
keithdoggett: Yes, adding that conditional will be good!

In the most recent commit, I have this as `lookup`

```rb
def lookup(*args, adapter: current_adapter_name, **kwargs)
  super(*args, adapter: current_adapter_name, **kwargs)
rescue ArgumentError => e
    if current_adapter_name == :postgis
     super(*args, adapter: :postgresql, **kwargs)
   else
     raise e
  end
end
```
Which works on all versions, but for some reason calling just `super` without the arguments causes errors on Ruby 3+. It looks like the call in the `rescue` block is still using `:postgis` as the adapter name for whatever reason. Adding the args to the first `super` call fixes it though. Have you seen anything like that before?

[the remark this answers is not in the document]
BuonOmo: I haven't seen that before, but I guess it is due to the new ruby way ton handle the last hash argument

BuonOmo (approved):
LGTM! Great work, thank you 💪

PR #335, opened 2021-02-16
Access configuration_hash using symbols

bb:
The previous code with the string 'database' does no longer work in current Rails versions.

Configuration hash uses symbols since Rails 6.1. See 
https://github.com/rails/rails/commit/ce9b197cc902ce8b72b7b3b5a3ec7e507fdba858


keithdoggett:
Thanks, @bb for the fix!

Issue #337, opened 2021-04-02
rails test broken if user is not postgres superuser on rails 6 (works on rails 5)


Lapizistik:
### Description
I try to use this gem on a fresh rails 6 project with an ordinary postgres database user (that has no superuser rights). 
I therefore add the postgis extension to the database via postgres template (see below). This works for development mode, but running `rails test` fails (sometimes the first run succeeds) with 
`DRb::DRbRemoteError: PG::InsufficientPrivilege: ERROR:  permission denied for table spatial_ref_sys` (sometimes also for table `ar_internal_metadata` instead). Stopping or killing `spring` does not help.

The setup described below **fails** with **rails 6.1.3** and **6.0.3.1**.
The identical setup **works** with **rails 5.2.5** without errors.
It also **works** for rails 6 if I use a superuser role.

This is my first time using activerecord-postgis-adapter and I am not an expert in how rails 6 runs tests (as far as I could see it creates at least 4 test databases using them in parallel?) so maybe I am doing something wrong?

Thanks for your work!

### Setup
ruby version: 2.5
OS: debian buster
postgresql: 11.11-0+deb10u1
postgis: 2.5.1+dfsg-1
#### postgres (psql)
First create a postgis template in postgres:
```
postgres=# CREATE DATABASE postgis_template;
postgres=# \c postgis_template
postgres=# CREATE EXTENSION postgis;
postgres=# CREATE EXTENSION postgis_topology;
postgres=# UPDATE pg_database SET datistemplate = TRUE WHERE datname = 'postgis_template';
```
And some user:
```
postgres=# CREATE USER testing with createdb;
postgres=# \password testing
```
#### rails (6.1.3)
Following the instructions in the README:
```
rails new test_6_1_3 --database=postgresql           # new rails project
cd test_6_1_3
edit config/database.yml                             # see below  
echo "gem 'activerecord-postgis-adapter'">>Gemfile   # add the gem
bundle install                                       # install it
rails generate controller Example index    # generate something to test
rails db:create                            # create the databases (gets postgis through postgis_template)
rails db:migrate
rails test                      # this one succeeds sometimes (could not find any deterministic behavior here)
rails test                      # this fails (with postgis as db adapter on rails 6)!
```
#### `config/database.yml`
```
default: &default
  adapter: postgis
  template: postgis_template
  encoding: unicode
  host: localhost
  username: testing
  password: ********
development:
  <<: *default
  database: test_6_1_3_development
test:
[...]

PR #338, opened 2021-04-16
Fix constants table truncation

joegaudet:
This commit addresses an issue where when you parallelize the rails
tests, it truncates all of the tables available through the connection
which in the case of PostGIS includes the spatial_ref_sys which is bad
news.

### DIFF
--- a/lib/active_record/connection_adapters/postgis/schema_statements.rb
+++ b/lib/active_record/connection_adapters/postgis/schema_statements.rb
@@ -35,6 +35,13 @@ module ActiveRecord
         end
 
         # override
+        # https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L51
+        # Returns an array of table names defined in the database, without the PostGIS spatial constants
+        def tables
+          query_values(data_source_sql(type: "BASE TABLE"), "SCHEMA").reject {|t| t == 'spatial_ref_sys'}
+        end
+
+        # override
         # https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb#L544
         #
         # returns Postgresql sql type string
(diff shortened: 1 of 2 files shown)


joegaudet:
Seems there are failures unrelated to my work.

keithdoggett:
Hi @joegaudet thanks for submitting this! This issue is closely related to #337 and I have a branch that fixes it (I just haven't opened a PR yet). The fix is in this commit (https://github.com/rgeo/activerecord-postgis-adapter/commit/074506fd1a4d19b455628845bc573c8d1ba59644) and you can build from the branch by adding `gem 'activerecord-postgis-adapter', git: 'https://github.com/rgeo/activerecord-postgis-adapter.git', branch: 'ignore-srs-truncate'` to your Gemfile.

We have different approaches to how we solved it, while yours is simpler I'm going to go with the fix on `ignore-srs-truncate` because it's narrower in scope. The proposed solution here would essentially hide the `spatial_ref_sys` table from ActiveRecord which may cause downstream issues and would probably be confusing to debug. Whereas the solution I proposed just hides it from the `trancate_tables` method which is what's causing the issue in the first place.

joegaudet:
Sounds good!

@keithdoggett have you also observed that when running parallelized tests that the table is not loaded when the db is forked.

I ended up copying in the setup SQL and running it if the table is empty.

Perhaps there's a better approach?

joegaudet:
@keithdoggett any thoughts?

keithdoggett:
@joegaudet I have not tried parallelized tests since I usually use rspec for my apps. Would you mind opening another issue for that and I can take a look?

keithdoggett:
Going to close this since #341 has been merged. If you're still seeing forking issues please open another issue and I can try to investigate. Thanks!

arthurwozniak:
Hi, is there any chance for new release with this fix in near time?

PR #341, opened 2021-04-26
Ignore Spatial Ref Sys Table during Truncate

### LINKED ISSUE #339: Spatial Constants Table Truncated when tests are running in parallel
Noticed when upgrading to rails 6.1 that this happens.

PR can be found here.

https://github.com/rgeo/activerecord-postgis-adapter/pull/338

keithdoggett:
Fixes #339 #337. During parallelized tests, ActiveRecord uses the `truncate_tables` method which truncates all of the tables in the database except for `ar_internal_metadata` and `schema_migrations`. This PR also excludes the `spatial_ref_sys` table from being truncated.

This is occurring on ActiveRecord 6+, so once approved this can be backported to 6-0-stable.


BuonOmo (approved):
LGTM

Issue #342, opened 2021-04-27
When using parallel tests out of the box post gis spatial ref table isn't copied over to test databases.


joegaudet:
Hey!

So I've observed that in addition to the truncation issue which you've kindly resolved. When an app is configured with activerecord-postgis-adapter and parallelized, the spatial ref table isn't copied over on initial setup. I worked around this by including the setup file and checking for its presence beforehand, but.... it's less than ideal.

I observed this using the latest rails + minitest with parallel test enabled as per the guides:

https://guides.rubyonrails.org/testing.html#parallel-testing

keithdoggett:
So I looked into this and the problem is that after the database is forked, it sets up the database with the `load_schema` method. Since we ignore all of the spatial tables they aren't copied over after the fork. 

Do you know if PostGIS is installed on the forked database? The fix might be as simple as adding an `after_fork_hook`.

joegaudet:
It is yes!

My temporary fix was to check-in the SQL file that populates that table and
load it if need be. But that felt a bit like a hack :P

.joe

joegaudet:
@keithdoggett this is the approach I used, but it required me to check-in the .SQL file which is not really great, is there a command I could run instead?

```
  parallelize_setup do |worker|
    # The parallel runners don't seem to always bring along the spatial table
    # If the spatial table has been nuked for some reason load it back up.
    if ActiveRecord::Base.connection.execute('select * from spatial_ref_sys').count == 0
      ActiveRecord::Base.connection.execute(IO.read("db/spatial_ref_sys.sql"))
    end
  end
```

keithdoggett:
Thanks that's helpful. I haven't had time to look into this more, but I think we may have to add a custom hook into the adapter that either installs PostGIS again to redo the setup or copies the `spatial_ref_sys` table over from the main database.

joegaudet:
Yeah makes sense, let me know if I can help anyway, I'm sorta blocked on this so happy to collaborate if you could point me in the right direction?

If there's a command that I could invoke that sets it up that'd probably be enough.

I'm not sure how I'd get a handle on the main table to copy it over tho.

keithdoggett:
[...]

PR #345, opened 2021-06-17
Include PostGIS::DatabaseStatements in Adapter

keithdoggett:
Fix for #342 and #337.

Instead of prepending the `DatabaseStatements` from #338, this includes it into the PostGIS adapter. Should fix issues where it would not be properly included in testing environments.

### DIFF
--- a/lib/active_record/connection_adapters/postgis/database_statements.rb
+++ b/lib/active_record/connection_adapters/postgis/database_statements.rb
@@ -2,12 +2,13 @@
 
 module ActiveRecord
   module ConnectionAdapters # :nodoc:
-    module PostGISDatabaseStatements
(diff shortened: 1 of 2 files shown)


BuonOmo (approved):
The fix is nice, however, the overriding of `truncate_tables` seems like we could get other issues on that from users trying to remove this table.

Anyway, lets wait until such an issue is opened before trying to find a more accurate way to avoid its deletion while letting a user the possibility to do it.

Could you consider adding a test in front of that change to avoid regression?

keithdoggett:
I think that luckily `truncate_tables` is invoked differently than `drop_tables` so users would still be able to drop the tables if need be. `truncate_tables` is used for testing setup.

Issue #313, opened 2020-03-31
Create ActiveRecord migration to enable postgis extension (if app is on Rails)


robinboening:
Hey 👋 

I'd like to suggest a feature to make the extension enabling process a bit more transparent.

Today we had an issue in our Rails application where a `CREATE EXTENSION` statement in the schema went missing within a PR. This issue is unrelated to postgis, byt we were confused so we checked our migrations to see what extensions we enabled. I got even more confused when I noticed that `postgis` wasn't even enabled by us, until my colleague pointed me to this gem and I found the `PostgisDatabaseTasks` enabling it when called by a rake task.

I'd like to suggest to not directly enable the extension by calling a rake task, but instead creating a migration file in which its handled. Of course, only if `Rails` is available. It makes it transparent.

Thanks for all your awesome work! 👌

seuros:
This rake task is legacy. it was implemented when rails did not have the ability to create extensions except by sql.

@keithdoggett i can push changes to remove it in V8.0.0 (as we cannot drop it for the current version)

woodardj:
If I'm reading this (and several other issues) right, the `rails db:gis:setup` script should not be used anymore, is that accurate? The Readme.md at the project root still lists it as an installation step.

woodardj:
Awesome, thanks @keithdoggett, makes sense. I ran into some weird connection errors when running that script against my Rails 6.1 + Postgres 13 project; I was about to file an issue for it, but I won't waste your time if that's not the future. (I was able to get up and running by writing a migration by hand)

seuros:
I removed the rake task and updated the README in #353. I think having the migration embedded will be an overkill , since this gem will need be a rails engine.

tindron:
> I removed the rake task and updated the README in #353. I think having the migration embedded will be an overkill , since this gem will need be a rails engine.

What do you think of adding this to History.md for 8.0 as a breaking change?  It was a little surprising to have my CI tests break after the upgrade due to not having the migration.

keithdoggett:
Added documentation around this in this History file and the README #368

PR #351, opened 2021-09-24
Support Activerecord 7.0

seuros:
Disabled jruby testing until it fixed upstream. 

Add docker-compose.yml to spin a postgis server for local development.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,9 +22,9 @@ jobs:
       BUNDLE_GEMFILE: ${{matrix.gemfile}}
     strategy:
       matrix:
-        ruby: ['3.0', '2.7', '2.6', '2.5', jruby]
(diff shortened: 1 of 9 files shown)


keithdoggett:
Hi @seuros thanks for the work on this! Soon I'm going to push a new branch for work on ActiveRecord 7 and I'll change the target branch of this PR to that branch. I'm thinking about making some broader changes to the testing suite where we test against ActiveRecord tests as well so we can try to mitigate issues down the line.

I'll do a full review of this shortly after I get the new branch up. Thanks!

keithdoggett:
Yes, that would be great thanks. I think that we should replace it with a rake task that will generate a migration that runs `enable_extension('postgis')`, exactly what is described in #313 . 

It should be fairly easy, but I'd like to have a way to enable it from the gem.

PR #352, opened 2021-10-18
ActiveRecord 7

keithdoggett:
Support for ActiveRecord 7

TODOs:

- [x]  Create stable branch of version 7
- [x]  ~~Create PostGIS extension migration for Rails apps~~
- [x] Test against ActiveRecord test suite (#358)


seuros:
The migration is described in the readme. I don't think we should init an engine just for a basic migration.

keithdoggett:
Agreed

seuros:
I will fix it. I will try to reproduce itl

keithdoggett:
Thanks for reporting that! Do we still need a Railtie even if we don't load our own tasks in order to have the Postgres Adapter's tasks?

seuros:
@keithdoggett No need. 
I had to register the tasks only. #355  has the fix.

test/attributes_test.rb
```diff
@@ -71,7 +71,7 @@ def test_spatial_attributes
 
     assert_equal 3857, data.polygon.srid
     assert_equal 3857, data.path.srid
-    assert_equal data.path, data.polygon.exterior_ring
+    assert_equal data.path.to_s, data.polygon.exterior_ring.to_s
```
keithdoggett, line 74: noting that this still needs to be looked at after merge

keithdoggett:
Version 8 has been released

PR #353, opened 2021-12-05
remove rake tasks from the adapter

seuros:
PS: We don't need Appraisal as we are not testing multiple active-record version anymore .


keithdoggett:
Thanks for working on this! Can you actually add the appraisal stuff back? Currently, we're only testing this against ActiveRecord 7.0, but when ActiveRecord 7.1 is released this version may be compatible and I'd like to have that infrastructure set up already.

seuros:
@keithdoggett From my experience working on ActiveRecord adapters, i think it better if we keep 8.x.x for rails 7.0 and we release 9.x for 7.1 because there will be changes in syntax . It a bad pattern to check the version on runtime for an adapter that is used 90% of the time.

We already have this logic in previous versions starting from activerecord 4.2 .

In 5.x we supported 2 versions, but you can see deprecation warning in the activerecord 5.2.

If you think that the setup still needed, i will push back the files.

Gemfile
```diff
@@ -5,5 +5,6 @@ gemspec
 
 gem "pg", "~> 1.0", platform: :ruby
 gem "activerecord-jdbcpostgresql-adapter", platform: :jruby
+gem "activerecord", "~> 7.0.0.alpha2"
```
keithdoggett, line 8: we should bump this to rc1 now.

test/attributes_test.rb
```diff
@@ -71,7 +71,7 @@ def test_spatial_attributes
 
     assert_equal 3857, data.polygon.srid
     assert_equal 3857, data.path.srid
-    assert_equal data.path, data.polygon.exterior_ring
+    assert_equal data.path.to_s, data.polygon.exterior_ring.to_s
```
keithdoggett, line 74: was there an issue with this?

test/basic_test.rb
```diff
@@ -105,7 +105,7 @@ def test_set_point_wkt_wrong_type
   def test_default_value
     create_model
     obj = SpatialModel.create
-    assert_equal factory.point(0, 0), obj.default_latlon
+    assert_equal factory.point(0, 0).to_s, obj.default_latlon.to_s
```
keithdoggett, line 108: same here. Some sort of object comparison issues?

keithdoggett (changes requested):
@seuros that's a fair point about the Appraisals and I agree we can just use more versions to properly handle each release. Especially since the releases now all seem to include fairly major changes.

Thank you for all the work here it looks very good for the most part I just have a few questions about some of the tests and want us to bump the AR version.

Gemfile
seuros: done

test/attributes_test.rb
seuros: yes,  the new release of rgeo is not using the same object. but it values are the same.

keithdoggett: hmm that seems weird. Do you know exactly what was causing the issue?

seuros: latest version of rgeo causing the issue. I will check that upstream then i will update this test later.

keithdoggett (approved):

keithdoggett:
Thanks @seuros

PR #355, opened 2021-12-13
cleanup

seuros:
Cleanup outdated documentation/config
Enable MFA for publishing the gem.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,5 +1,11 @@
 name: Tests
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
 
 jobs:
   test-ubuntu-ruby:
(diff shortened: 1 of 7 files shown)


.github/workflows/tests.yml
```diff
@@ -1,5 +1,11 @@
 name: Tests
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
keithdoggett, line 9: We need to run tests on all PRs

keithdoggett (changes requested):

.github/workflows/tests.yml
seuros: Yeah . This make sense.

keithdoggett: The block should look like this then

```
on:
  push:
    branches:
      - master
  pull_request:
```

Right now it's only running PRs on branches that are merging with `master` I believe. Tests didn't run on this PR yet.

seuros:
@keithdoggett  this PR fix the tasks issue too.

keithdoggett (approved):
LGTM thank you for the continued work on this

PR #356, opened 2021-12-17
Updated to support activerecord 7

DukeCityDev:


DukeCityDev:
I know this is really half-assed, maybe I should have just opened an issue. Apologies! Either way so far it's working for me with just changing the gemspec. If there's anything else I can do to help let me know!

PR #360, opened 2022-01-11
Port recent changes to PostgreSQL adapter from Rails 7

ddiachkov:
Rails team recently added a support for generated columns and fixed a couple of bugs in the adapter. This PR ports those changes to the activerecord-postgis-adapter.


lib/active_record/connection_adapters/postgis/schema_statements.rb
```diff
@@ -48,29 +48,41 @@ def new_column_from_field(table_name, field)
...
+              end
+            when "integer"
+              case limit
+              when 1, 2; "smallint"
+              when nil, 3, 4; "integer"
+              when 5..8; "bigint"
+              else raise ArgumentError, "No integer type has byte size #{limit}. Use a numeric with scale 0 instead."
+              end
+            when "enum"
+              raise ArgumentError, "enum_type is required for enums" if enum_type.nil?
```
keithdoggett, line 77: Are these not handled when we called `super` originally? Since the adapter subclasses the `PostgresqlAdapter` I don't think we have to duplicate this.

keithdoggett:
Thanks for submitting this. The only thing I'm curious about is if we need to duplicate that logic in that comment I left. Otherwise looks great.

lib/active_record/connection_adapters/postgis/schema_statements.rb
ddiachkov: You're correct! I missed the fact that SchemaStatements is mixed-in into ConnectionAdapter.
It looks like we don't need `type_to_sql` at all. This code https://github.com/rails/rails/blob/829427306c7bb2dfff47ac5fddcd066bb2e0a063/activerecord/lib/active_record/connection_adapters/abstract/schema_statements.rb#L1252 generates valid SQL as long as type is registered in `native_database_types`.

keithdoggett: Good catch! That's nice we can just fall back on that

keithdoggett:
LGTM! I'm going to let @BuonOmo take a look and then can merge soon

lib/active_record/connection_adapters/postgis/schema_statements.rb
```diff
@@ -5,18 +5,17 @@ module ConnectionAdapters
...
           default_value = extract_value_from_default(default)
           default_function = extract_default_function(default_value, default)
 
-          serial =
-            if (match = default_function&.match(/\Anextval\('"?(?<sequence_name>.+_(?<suffix>seq\d*))"?'::regclass\)\z/))
-              sequence_name_from_parts(table_name, column_name, match[:suffix]) == match[:sequence_name]
-            end
+          if match = default_function&.match(/\Anextval\('"?(?<sequence_name>.+_(?<suffix>seq\d*))"?'::regclass\)\z/)
+            serial = sequence_name_from_parts(table_name, column_name, match[:suffix]) == match[:sequence_name]
+          end
```
BuonOmo, line 18: I'd rather not include cosmetic changes in meaningful (eg not fully cosmetic) PRs, this only make reviewing and blaming harder unfortunately, plus that is opiniated (for instance, I favor parenthis in if conditions with assignment to distinguish those).. Could you revert that bit please?

BuonOmo (approved):
Thanks for contributing 🙂 

Apart from that one unnecessary cosmetic change (I guessed, after reviewing a few times), LGTM!

@keithdoggett shouldn't we add this to the contributing guidelines?

lib/active_record/connection_adapters/postgis/schema_statements.rb
ddiachkov: This is a verbatim copy from: https://github.com/rails/rails/blob/ba8d029e0c9965f64c96f44326fe40f0df6293f3/activerecord/lib/active_record/connection_adapters/postgresql/schema_statements.rb#L668-L670

The change is cosmetic but I think it is important to keep this method as close as possible to the Rails codebase since we're overwriting it.  Should I revert it still?

BuonOmo: Oh yeah makes definitely more sense then to let it asis, thanks for the explanation :)

PR #363, opened 2022-03-28
Test Against ActiveRecord

### LINKED ISSUE #358: Test Against ActiveRecord
We should test the adapter against the full suite of ActiveRecord tests to assure we have not inadvertently caused any errors.

keithdoggett:
Fixes #358 

Add functionality to test the PostGIS adapter against the full ActiveRecord test suite for Postgres.

- Moved all tests to `test/cases` and namespaced them with `module PostGIS` to avoid potential conflicts.
- Added a script to the `Gemfile` that pulls the matching Rails version to the ActiveRecord version specified in the `gemspec` so we can access the tests.
- Added some methods to `rake_helper` that help with getting the libs and tests from ActiveRecord.
- Changed `test/database.yml` to be in an ActiveRecord test compatible format.
- Add `TestTask`s to the `Rakefile` that allow testing on just local PostGIS tests, ActiveRecord tests, or both. By default it is only local tests.
  - `bundle exec rake test:postgis`
  - `bundle exec rake test:activerecord`
  - `bundle exec rake test:all`

TODOs:
- [ ] Figure out way to skip specific ActiveRecord tests if necessary
- [ ] Open issue to request help with ActiveRecord compatibility
- [ ] Have way to track ActiveRecord compatibility automatically?


.gitignore
```diff
@@ -8,3 +8,5 @@ Gemfile.lock
 /travis/*.lock
 /test/database_local.yml
 .idea
+debug.log
+test/db/*
```
BuonOmo, line 12:
```suggestion
/test/db/*
```

This makes it explicit that the path only comes from project root :)

CONTRIBUTING.md
```diff
@@ -37,6 +37,22 @@ Make sure the tests pass:
...
+
+`POSTGIS_TEST_FILES=test/cases/ddl_test.rb bundle exec rake`
+
+If you are testing a feature against the ActiveRecord test suite run:
+
+`bundle exec rake test:activerecord`
+
+Files can be specified with the `AR_TEST_FILES` environment variable:
+
+`AR_TEST_FILES test/cases/adapters/postgresql/*_test.rb bundle exec rake test:activerecord`
```
BuonOmo, line 50:
```suggestion
`AR_TEST_FILES=test/cases/adapters/postgresql/*_test.rb bundle exec rake test:activerecord`
```

Gemfile
```diff
@@ -7,3 +7,34 @@ gem "pg", "~> 1.0", platform: :ruby
 gem "activerecord-jdbcpostgresql-adapter", platform: :jruby
 gem "ffi-geos", platform: :jruby
 gem "byebug" if ENV["BYEBUG"]
+
+def activerecord_version
```
BuonOmo, line 11: This gets only one version, shouldn't we test against multiple rails versions ? For that, we could do as with ruby: take advantage of the test matrix in github ci.

In that case the version could simply be an env variable or a task argument

test/test_helper.rb
```diff
@@ -1,18 +1,41 @@
...
+  def load_postgis_specific_schema
+    original_stdout = $stdout
+    $stdout = StringIO.new
+
+    load "schema/postgis_specific_schema.rb"
+
+    ActiveRecord::FixtureSet.reset_cache
+  ensure
+    $stdout = original_stdout
+  end
```
BuonOmo, line 26: Why this stdout stuff?

test/rake_helper.rb
```diff
@@ -0,0 +1,52 @@
...
+def set_arconfig_env!
+  arconfig_file = File.expand_path(File.join(File.dirname(__FILE__), "database.yml"))
+  ENV["ARCONFIG"] = arconfig_file
+end
+
+config_load_paths!
+set_arconfig_env!
+
+def activerecord_test_files
+  if ENV["AR_TEST_FILES"]
```
BuonOmo, line 26: why not a single `TEST_FILES` constant to have only one thing to remember ? :)

test/rake_helper.rb
```diff
@@ -0,0 +1,52 @@
...
+def postgis_test_load_paths
+  ["lib", "test", File.join(ar_root, "lib"), File.join(ar_root, "test")]
+end
+
+def config_load_paths!
+  postgis_test_load_paths.each { |p| $LOAD_PATH.unshift(p) unless $LOAD_PATH.include?(p) }
+end
+
+def set_arconfig_env!
+  arconfig_file = File.expand_path(File.join(File.dirname(__FILE__), "database.yml"))
```
BuonOmo, line 18:
```suggestion
  arconfig_file = File.expand_path("database.yml", __dir__)
```

BuonOmo (commented):
Looks good overall, however I think using a test matrix would make a bit more sense, and there are some simplifications available I think :)

test/rake_helper.rb
keithdoggett: It's possible that we will have conflicting files which could be confusing for a user. For example `test/cases/type_test.rb` is defined in both postgis and ActiveRecord so it would run 2 test files if we only had one environment variable. Maybe it's not a big deal though.

Gemfile
BuonOmo: I'm not certain why having both is mandatory. Actually I think I am missing the point of this method :/

keithdoggett: Gotcha. The point of this method is that for local development if we want to test against ActiveRecord, we need to add Rails to the bundle and fetch it from its git repo in order to access all of the tests. If we just use the ActiveRecord from Rubygems the test directory is not included, same with Rails from Rubygems.

So this method reads in our gemspec, finds what ActiveRecord version we've specified as a dependency, fetches all of the Rails releases from Rubygems, finds the highest matching version (i.e. ActiveRecord ~> 7.0.0 resolves to ActiveRecord=7.0.2.3 currently), and then pulls the corresponding Rails version from git since all releases are tagged on git as well.

Gemfile
```diff
@@ -7,3 +7,36 @@ gem "pg", "~> 1.0", platform: :ruby
...
+  ver = versions.find { |v| ar_dep.match?("activerecord", v["number"]) }
+
+  raise Bundler::GemNotFound, "No matching Activerecord version found for #{ar_dep.requirement}" unless ver
+
+  ver["number"]
+end
+
+group :development do
+  # Need to install for tests
+  gem "rails", git: "https://github.com/rails/rails.git", tag: "v#{activerecord_version}"
```
BuonOmo, line 36:
```suggestion
  gem "rails", github: "rails/rails", tag: "v#{activerecord_version}"
```

YAGNI but sexy ^^

BuonOmo (approved):
LGTM!

keithdoggett:
Some ActiveRecord specific tests are still failing, but we can handle those in separate issues/PRs as most of the remaining failures will require us to override/re-write the original test from Rails.

PR #373, opened 2022-09-13
Force use of geos based wkb parsing

AndrewHYi:
From upstream:

```
Version 8.x supports ActiveRecord 7.0
Requirements:

ActiveRecord 7.0
Ruby 2.7.0+
PostGIS 2.0+
```

PR #376, opened 2022-11-28
Test against RGeo3

keithdoggett:
I had to change a few test cases to work with the GEOS WKT generator. The library shouldn't have needed any changes since this doesn't directly depend on `rgeo` and only depends on `rgeo-activerecord`, but it's probably a good idea to test anyways... and it doesn't need any changes.

I added `rgeo` to the gemfile since bundler won't install prerelase versions without explicitly setting them somewhere, but I will remove that once version 3 is released.


test/basic_test.rb
```diff
@@ -196,11 +196,11 @@ def test_multi_polygon_column
     end
     SpatialModel.reset_column_information
     rec = SpatialModel.new
-    wkt = "MULTIPOLYGON (((-73.97210545302842 40.782991711401195, " \
+    wkt = "MULTIPOLYGON (((-73.97210545302842 40.78299171140119, " \
```
BuonOmo, line 199: Why this change ?

keithdoggett: I'm not sure. For whatever reason the 15th decimal in this wkt was getting truncated on the output. I tried looking into it a bit, but I'm not sure if it's a PostGIS change or something to do with the WKT parsers in RGeo. I'll see if I can figure out exactly why it's failing.

Also kind of related but this PR (https://github.com/rgeo/activerecord-postgis-adapter/pull/373) pointed out that we use the Ruby based parsers here. I can open a separate PR that uses the Geos based parsers if available.

keithdoggett: @BuonOmo it works now so i'm not sure what was going on before.

Issue #369, opened 2022-08-01
Does not dump virtual / generated / stored columns correctly


sfcgeorge:
Whatever they're called, when using the postgis adapter with Rails 7 virtual columns don't get dumped correctly leading to an invalid schema.

```diff
- t.virtual "popularity", type: :string, as: "1", stored: true
+ t.string "popularity", default: -> { "1" }
```

keithdoggett:
I just released version 8.0.1 which includes this PR (https://github.com/rgeo/activerecord-postgis-adapter/pull/360) that I believe resolves this issue. Let me know if it doesn't.

Tristramg:
Hello, I’m having some troubles specifically with virtual geometry columns in the generated schema.rb

The migration works fine with `limit: "geometry, 3857"`

But I get something like
`t.virtual "geom_web_mercator", type: :geometry, limit: {:srid=>0, :type=>"geometry"}, as: "st_transform(geom, 3857)", stored: true`

So when doing a `rails db:reset`  I have the following error:

```
ActiveRecord::StatementInvalid: PG::SyntaxError: ERROR:  syntax error at or near "{"
LINE 1: ...:geography)) STORED, "geom_web_mercator" geometry({:srid=>38...
```

PR #377, opened 2023-04-21
overload type_to_sql

Tristramg:
This fixes when using generated geometry column from schema.rb

ref #369

`type_to_sql` had been removed in https://github.com/rgeo/activerecord-postgis-adapter/pull/360/files#diff-eeeb5e84f6023e4b440660467863dbed44712213c7c65db4e6fe401a3c219b1cL37-L57 as the adapter knows how to handle limits as strings. But in this situation, it is as a hash

### DIFF
--- a/lib/active_record/connection_adapters/postgis/spatial_table_definition.rb
+++ b/lib/active_record/connection_adapters/postgis/spatial_table_definition.rb
@@ -8,7 +8,13 @@ module ActiveRecord  # :nodoc:
 
         # super: https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb
         def new_column_definition(name, type, **options)
-          if (info = PostGISAdapter.spatial_column_options(type.to_sym))
+          col_type = if type.to_sym == :virtual
+            options[:type]
+          else
+            type
+          end
+
+          if (info = PostGISAdapter.spatial_column_options(col_type))
             if (limit = options.delete(:limit)) && limit.is_a?(::Hash)
               options.merge!(limit)
             end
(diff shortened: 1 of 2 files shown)


keithdoggett:
Hey @Tristramg thanks for submitting this PR. I think it looks good, but we should add a test to make sure. I think we may need to add more types to that array as well in case people make columns as `t.st_polygon` for example.

Here's a sample test that we can add to test/ddl_test.rb

```ruby
  def test_generated_geometry_column
    klass.connection.create_table(:spatial_models, force: true) do |t|
      t.st_point :coordinates, limit: { srid: 4326 }
      t.virtual :generated_buffer, type: :st_polygon, limit: { srid: 4326 }, as: 'ST_Buffer(coordinates, 10)', stored: true
    end
    klass.reset_column_information
    col = klass.columns.last
    assert_equal(:geometry, col.type)
    assert(col.virtual?)
  end
```

keithdoggett:
Looking into this more, I think we need to modify `new_column_definition` in `spatial_table_definition.rb` and check if a column with the type `:virtual` is also a spatial type. Then what you have should work.

Tristramg:
Thank you a lot. You basically wrote the whole commit !

Much better in `new_column_definition` as it avoids doing the same work twice

Tristramg:
Thank you for the reminder. I’ve rebased my fork

keithdoggett:
Looks like it's failing with older versions of postgres since it is probably not supported before v12. We can add a `skip` statement at the beginning of it with a conditional to check for virtual columns being supported.

`skip "Virtual Columns are not supported in this version of PostGIS" unless SpatialModel.connection.supports_virtual_columns?`

Tristramg:
Let’s see if this time is the right one. Thank you for you patience and walking me through all this :)

keithdoggett:
No problem! Thanks for the contribution! I can get a release out in the next day or 2 with this change.

PR #379, opened 2023-06-21
[CI Update] Ruby 3.2 and PG 15

berkos:
👋  This PR tests against Ruby 3.2 and Postgres 15
Thanks for the review in advance 🙏

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,8 +27,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


BuonOmo:
@berkos what do you say, to avoid running machines uselessly we get rid of versions past eol?

- ruby before 3.0 https://www.ruby-lang.org/en/downloads/branches/
- pg before 11 https://www.postgresql.org/support/versioning/

BuonOmo:
@berkos LGTM, thank you for your contribution ! 🌮

berkos:
Thank you for the quick review! Good point about the EOL versions.

PR #380, opened 2023-08-13
Loosen activerecord dependency to allow rails 7.1.0.alpha

manastyretskyi:
It's impossible to install gem in rails app that is running on rails master right now, so I updated dependency similarly to [rgeo-activerecord](https://github.com/rgeo/rgeo-activerecord/blob/80edc88f649f6ec8488990cb99ddec0c7153091b/rgeo-activerecord.gemspec#L17), so that get will be compatible with rails 7.1

### DIFF
--- a/activerecord-postgis-adapter.gemspec
+++ b/activerecord-postgis-adapter.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.7.0"
 
-  spec.add_dependency "activerecord", "~> 7.0.0"
+  spec.add_dependency "activerecord", ">= 7.0.0"
   spec.add_dependency "rgeo-activerecord", "~> 7.0.0"
 
   spec.add_development_dependency "rake", "~> 13.0"


manastyretskyi:
I've realised that there are some changes in ActiveRecord 7.1 that will require more work to make gem work with it, so I will close this PR for now and will re-open it after rails 7.1 release, in case it will not be done by that time by someone else.

BuonOmo:
May you point out what changes you are talking about ? It would be interesting for us maintainers.

manastyretskyi:
@BuonOmo You can see changes that I did [here](https://github.com/rgeo/activerecord-postgis-adapter/compare/master...kadastrova-karta:activerecord-postgis-adapter:master), it not sure if that is all that is needed, but at least tests are passing.

keithdoggett:
@BuonOmo we should open up a separate branch that is built against this version of Rails. Depending on how many changes there are we may be able to eventually merge it back into this major version or we'll have to release another major version. The last time there was a minor release for Rails it caused us to make backwards incompatible changes so I generally think it's best to keep these changes away from main until the new version is officially released.

BuonOmo:
Fully on page with you @keithdoggett. I don't think we are in any rush yet anyway!

and @manastyretskyi if you want to open up a PR doing this that would be very welcome :)

PR #385, opened 2023-10-12
Pass new identity attribute to preserve argument order

joseramonc:
https://github.com/rails/rails/pull/49504 introduced a new `identity` argument on the `ActiveRecord::ConnectionAdapters::PostgreSQL::Column#initialize` method which is causing the dumping of virtual columns to be treated as identity columns. 

This PR updates the methods to match the new ones on the Postgres adapter.


keithdoggett:
Thanks @joseramonc LGTM!

Issue #389, opened 2023-12-01
Rails 7.1 db:create not installing extension


frehoy:
On a fresh new rails app on Rails 7.1.2 with the gem installed and `adapter: postgis` in config/database.yml I run `rake db:create` but the postgis extension is not created in the db, but the readme implies it will be: 

> The adapter will add the PostGIS extension to your database.
https://github.com/rgeo/activerecord-postgis-adapter#creating-a-spatial-rails-app 

Running a migration with `enable_extension 'postgis'` works fine though. 

I'd be happy to fix the readme or take a stab at fixing it in code if someone can point me in the right direction, I can't find the code that's supposed to create the extension 😬

BuonOmo:
I don’t have access to my computer now, I would find the correct place by following the db create task, found on GitHub: https://github.com/rails/rails/blob/531643244474446efc061b559504ed67acc8cae1/activerecord/lib/active_record/railties/databases.rake#L46C14-L46C14

BuonOmo:
@frehoy are you still in for fixing this ?

I think I found the culprit commit: https://github.com/rgeo/activerecord-postgis-adapter/pull/353/commits/418c02706291a595ac4e748d18f5c64459444ad2

The author removed loads of things, including the `postgis_database_tasks.rb` that contained an `extension_names` method. Which is the interesting bit for us. Based on this I think there is enough to implement a PR (of course, we'd need to find which are the good bits we want). I don't really have time or energy for this now unfortunately so your help would be very appreciated !

frehoy:
Yes! I'll give it a go this week.

frehoy:
Thanks for the pointers @BuonOmo, I had a look at https://github.com/rgeo/activerecord-postgis-adapter/commit/418c02706291a595ac4e748d18f5c64459444ad2 now and found quite a few tasks. 

Since I don't know the background of why the tasks were removed I think just making the README consistent is the best course of action right now? I made a PR for that. Adding automagically creating the extensions back seems like a fair bit of work and I'm not sure of how it "should" look. Here's the PR: https://github.com/rgeo/activerecord-postgis-adapter/pull/392

frehoy:
Fixed by https://github.com/rgeo/activerecord-postgis-adapter/pull/392 Thanks for the help!

PR #392, opened 2024-02-03
Clarify extension creation instructions in README

frehoy:
The "Creating a Spatial Rails App" previously said "The adapter will add the PostGIS extension to your database.", which isn't true since this commit: https://github.com/rgeo/activerecord-postgis-adapter/commit/418c02706291a595ac4e748d18f5c64459444ad2

I updated the readme instructions to add the migration to create the extension in the "Creating a Spatial Rails App" section. It was then almost identical to the "Upgrading an Existing Database" section, so I merged the two into a new section called "Setup". 

I also went for the newer style `default: &default` style for the `config/database.yml` part. 

Fixes issue https://github.com/rgeo/activerecord-postgis-adapter/issues/389

Thanks @BuonOmo for the pointers

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,38 +222,39 @@ development:
 This adapter uses the `rgeo` gem, which has additional dependencies.
 Please see the README documentation for `rgeo` for more information: https://github.com/rgeo/rgeo
 
-## Creating a Spatial Rails App
+## Setup
 
-This section covers starting a new Rails application from scratch. If you need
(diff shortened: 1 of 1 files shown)


BuonOmo (approved):
Thanks for the update @frehoy.

@keithdoggett WDYT?

keithdoggett (approved):

keithdoggett:
Thanks @frehoy LGTM

PR #396, opened 2024-06-04
Rails 7.2

ugisozols:


keithdoggett:
Looks like we'll have to bump the minimum ruby version to `>= 3.1.0`

seuros:
We need to use the new api in active-record so it doesn't blow up when we add other gems. 

I remember i had flaky tests when i tried to do it few months ago.

BuonOmo:
We should close either this PR or the other ! I can champion this in a few weeks if no one has been on it by that time. @ugisozols still interested in going forward with this PR?

ugisozols:
Sorry, I got busy with other stuff so let's close this and better focus on the other one.

PR #397, opened 2024-06-18
feat!: ActiveRecord 7.2 support

### LINKED ISSUE #395: Inquiry on Rails 7.2 Support Effort
Hi everyone,

I wanted to check if anyone is currently working on adding support for Rails 7.2. If there is already an effort underway, I would love to know more about it and see if there's any way I can assist.

If no one is handling this task, I am planning to dedicate some time over the weekend to work on it. 

Remember this will be again a breaking change since the new version will support just 7.2.x

seuros:
Here is a working implementation for Rails 7.2

ping @theonlyriddle @anthony0030 @ugisozols @keithdoggett 

JRuby code was removed until it start supporting Rails 7.1 and 7.2

closes #395


seuros:
@keithdoggett : The errors are flacky and happen when the adapter is not connected yet in CI. 
rerunning the test will make them pass.

lib/activerecord-postgis-adapter.rb
```diff
@@ -1,3 +1,5 @@
 # frozen_string_literal: true
 
-require_relative "active_record/connection_adapters/postgis_adapter"
+require 'active_record'
+require "active_record/connection_adapters"
+ActiveRecord::ConnectionAdapters.register("postgis", "ActiveRecord::ConnectionAdapters::PostGISAdapter", "active_record/connection_adapters/postgis_adapter")
```
seuros, line 5: this will lazy load the adapter until it used.

test/cases/basic_test.rb
```diff
@@ -180,7 +180,7 @@ def test_point_to_json
       obj = SpatialModel.new
       assert_match(/"latlon":null/, obj.to_json)
       obj.latlon = factory.point(1.0, 2.0)
-      assert_match(/"latlon":"POINT\s\(1\s2\)"/, obj.to_json)
+      assert_match(/"latlon":"POINT\s\(1(\.0)?\s2(\.0)?\)"/, obj.to_json)
```
seuros, line 183: Some versions return a full float when with the digit is 0.

test/cases/schema_statements_test.rb
```diff
@@ -5,22 +5,25 @@
 module PostGIS
   class SchemaStatementsTest < ActiveSupport::TestCase
     def test_initialize_type_map
-      initialized_types = SpatialModel.connection.send(:type_map).keys
+      SpatialModel.with_connection do |connection|
+        assert connection.connected?
```
seuros, line 9: This is flacky.

lib/active_record/connection_adapters/postgis/create_connection.rb
```diff
@@ -1,50 +1,27 @@
...
-          logger,
-          conn_params,
-          config
-        )
-      end
-
+    # Based on the default <tt>postgresql_connection</tt> definition from ActiveRecord.
+    # https://github.com/rails/rails/blob/master/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb
+    # FULL REPLACEMENT because we need to create a different class.
+    def postgis_connection(config)
```
seuros, line 8: I think we can merge this adapter into postgresql and enable extra types if the extension is detected.

We can do it in a future version.

.github/workflows/tests.yml
```diff
@@ -27,8 +27,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ruby-head, '3.2', '3.1', '3.0']
-        pg: [11-3.0, 12-master, 13-master, 14-master, 15-master]
+        ruby: [ruby-head, '3.3', '3.2', '3.1']
+        pg: [11-3.0, 12-master, 13-master, 14-master, 15-master, 16-master]
```
seuros, line 31: added test for pg 16.

keithdoggett:
@seuros thank you for submitting this I'll review this in the morning. Thanks for all the help.

anthony0030:
@seuros thank you for your continued effort. I and @theonlyriddle will be unable to work on the project that we need this for, for some time.

We appreciate your effort very much!

Gemfile
```diff
@@ -4,39 +4,13 @@ source "https://rubygems.org"
...
-
-  uri = URI("https://rubygems.org/api/v1/versions/activerecord.yaml")
-  res = Net::HTTP.get_response(uri)
-  versions = YAML.safe_load(res.body)
-  ver = versions.find { |v| ar_dep.match?("activerecord", v["number"]) }
-
-  raise Bundler::GemNotFound, "No matching Activerecord version found for #{ar_dep.requirement}" unless ver
-
-  ver["number"]
-end
```
keithdoggett, line 32: Any reason we're removing this? This is required in order to test against the full activerecord test suite. I know it's not part of the full CI yet, but IMO it's helpful to be able to run locally and potentially flag any issues.

lib/active_record/connection_adapters/postgis/create_connection.rb
keithdoggett: So that would mean fully subclassing the postgresql adapter and just augmenting the type_map and adding a spatial column?

test/cases/basic_test.rb
keithdoggett: yes, seen this before thanks for that

test/cases/schema_statements_test.rb
keithdoggett: Probably due to how the adapter is registered now?

test/rake_helper.rb
```diff
@@ -3,7 +3,7 @@
 POSTGIS_TEST_HELPER = "test/test_helper.rb"
 
 def ar_root
-  File.join(Gem.loaded_specs["rails"].full_gem_path, "activerecord")
+  Gem.loaded_specs["activerecord"].full_gem_path
```
keithdoggett, line 6: Does this allow the activerecord tests to get loaded in without installing the gem directly?

keithdoggett (approved):
@seuros looks good. Just a few questions here. Thanks again

Gemfile
seuros: We still pulling, just using a static tag instead.

test/rake_helper.rb
seuros: This is just a direct reference to the actual gem folder.

test/cases/schema_statements_test.rb
seuros: No, it just flacky in CI. 

I will focus on the cause later.

lib/active_record/connection_adapters/postgis/create_connection.rb
seuros: Yes. 
But let do it after this upgrade, after we test the impact.

test/cases/schema_statements_test.rb
BuonOmo: Was it flaky before this PR?

BuonOmo (commented):
A few reviews, to me the `Gemfile` part is important. I also have doubts about removing the JRUBY code, is it really bothering us? My opinion is that support will come at some point and that we will just have to write the same code again...

Besides that I think we're pretty good. Thank you for the dedication !

lib/active_record/connection_adapters/postgis/create_connection.rb
seuros: Postgis is just an extension in postgresql.
When we add vector or age extension, we dont switch the adapter.

anyway this can be discussed later.

Gemfile
seuros: This was needed when the adapter was supporting multiple versions (prior 5.1).
The CI never had a AR_VERSION set.

test/cases/schema_statements_test.rb
seuros: Yes, in CI.

Gemfile
BuonOmo: I'm not talking about AR_VERSION but the dynamic part below (fetching the latest ar patch!)

BuonOmo:
I think there is still some misunderstanding in the Gemfile part, to me it is important that we test on the latest AR patch.

seuros:
@BuonOmo ,  I get it. 
I want to lock it to ˜> 7.2.0 when we release it .

BuonOmo:
So the script you removed allowed to verify that we were properly testing against the latest patch. What you are doing isn't, unless I missed something which you could explain to me ?

seuros:
No the scrip that i removed, pull for each run the list from rubygems , select the last one and use it a tag.

You can bypass the network calls by setting AR_VERSION

BuonOmo:
> No the script that i removed, pull in each run the list from rubygems , select the last one and use it as a tag.

Which means it forces usage of the very last patch, which is what want IMHO!

seuros:
fair! I restored the file.

theonlyriddle:
@seuros I updated my gem to use your current master branch to see how things are running. When I started my application, it crashed at my initializer file `config/initializers/rgeo.rb` with the error `uninitialized constant RGeo::ActiveRecord (NameError)`.

The contents of the initializer file are below.

```
RGeo::ActiveRecord::SpatialFactoryStore.instance.tap do |config|
  # By default, use the GEOS implementation for spatial columns.
  # config.default = RGeo::Geos.factory_generator

  # # But use a geographic implementation for point columns.
  # config.register(RGeo::Geographic.spherical_factory(srid: 4326), geo_type: "point")
end

class RGeo::GeoJSON::Feature
  def properties
    @properties
  end
end
```

I only have this problem on your branch. Just thought to bring this to your attention.

seuros:
@theonlyriddle 
```
RGeo::ActiveRecord::SpatialFactoryStore.instance.tap do |config|
  # By default, use the GEOS implementation for spatial columns.
  # config.default = RGeo::Geos.factory_generator

  # # But use a geographic implementation for point columns.
  # config.register(RGeo::Geographic.spherical_factory(srid: 4326), geo_type: "point")
end
```

Is doing nothing in your case.

theonlyriddle:
@seuros I agree with you, the code I posted does nothing as it's commented out.

However, this code from the documentation (essentially the same as what I posted) results in the same error that I described above.

```
RGeo::ActiveRecord::SpatialFactoryStore.instance.tap do |config|
  # By default, use the GEOS implementation for spatial columns.
  config.default = RGeo::Geos.factory_generator

  # But use a geographic implementation for point columns.
  config.register(RGeo::Geographic.spherical_factory(srid: 4326), geo_type: "point")
end
```

Gemfile
```diff
@@ -4,39 +4,36 @@ source "https://rubygems.org"
 gemspec
 
 gem "pg", "~> 1.0", platform: :ruby
-gem "activerecord-jdbcpostgresql-adapter", platform: :jruby
-gem "ffi-geos", platform: :jruby
 gem "byebug" if ENV["BYEBUG"]
 
 def activerecord_version
-  return ENV["AR_VERSION"] if ENV["AR_VERSION"]
+   return ENV["AR_VERSION"] if ENV["AR_VERSION"]
```
BuonOmo, line 10: there's a slight indentation issue ^^

seuros:
@theonlyriddle, Right! 
we need to upgrade the documentation to configure it after initialization. 
Now the adapters are lazy loaded and available only after first usage.

BuonOmo (approved):
LGTM, I think once @theonlyriddle's issue is sorted out we should be good to go!

seuros:
@BuonOmo :shipit:

PR #403, opened 2024-08-21
Added rails 7.2.0

StoneGod:


BuonOmo:
@StoneGod do you plan on moving on with this PR or shall we look for an other champion ? :)

StoneGod:
Better look for an other champion

Issue #402, opened 2024-08-12
Error `TypeError: can't cast RGeo::Geographic::SphericalPolygonImpl` on Rails 7.2 and HEAD `activerecord-postgis-adapter`


stevegeek:
The following reproduction script shows the error. 

It works on Rails 7.0/7.1 with the latest release of `activerecord-postgis-adapter` but fails when run with Rails 7.2 and master branch of `activerecord-postgis-adapter`

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails", "~> 7.2"
  # If you want to test against edge Rails replace the previous line with this:
  # gem "rails", github: "rails/rails", branch: "main"

  gem "pg"
  gem "activerecord-postgis-adapter", github: "rgeo/activerecord-postgis-adapter", branch: "master"
  gem "rgeo-activerecord", "~> 7"
end

require "active_record"
require "minitest/autorun"
require "logger"


# `createdb postgis_test_db`

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(
  "postgis://localhost/postgis_test_db"
)
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  enable_extension 'postgis'

  create_table :posts, force: true do |t|
    t.st_point :geo
  end
end

class Post < ActiveRecord::Base
end

class BugTest < Minitest::Test
  def test_geo_where
    post = Post.create!(geo: RGeo::Geos.factory(srid: 4326).point(-122, 47))
    point = post.geo
    geo_factory = RGeo::Geographic.spherical_factory(srid: 4326, buffer_resolution: 4)
    area_center = geo_factory.point(*point.coordinates)
    circle = area_center.buffer(10_000)
    polygon = geo_factory.multi_polygon([circle])
    assert_equal 1, Post.where("ST_DWithin(?, posts.geo, ?)", polygon, 1).count
  end
end
```

Exception:

```
TypeError: can't cast RGeo::Geographic::SphericalPolygonImpl
    /Users/stephen/.asdf/installs/ruby/3.3.4/lib/ruby/gems/3.3.0/gems/activerecord-7.2.0/lib/active_record/connection_adapters/abstract/quoting.rb:105:in `type_cast'
[...]

PR #404, opened 2024-09-02
fix: Rails 7.2 can't cast error.

vicocamacho:
As part of the updates made to the PostgreSQL adapter for Rails 7.2. Rails changed the argument order for instantiating column objects here https://github.com/rails/rails/blob/a11f0a63673d274c59c69c2688c63ba303b86193/activerecord/lib/active_record/connection_adapters/postgresql/column.rb#L9. Because the PostGIS adapter inherits from the PosgreSQL adapter it requires just a small update to the order of the arguments when instantiating new columns.

It should fix the cast errors reported in https://github.com/rgeo/activerecord-postgis-adapter/issues/402

### DIFF
--- a/lib/active_record/connection_adapters/postgis/spatial_column.rb
+++ b/lib/active_record/connection_adapters/postgis/spatial_column.rb
@@ -31,7 +31,7 @@ module ActiveRecord  # :nodoc:
             build_from_sql_type(sql_type_metadata.sql_type)
           end
           super(name, default, sql_type_metadata, null, default_function,
-                collation: collation, comment: comment, serial: serial, generated: generated, identity: identity)
+                collation: collation, comment: comment, serial: serial, identity: identity, generated: generated)
           if spatial? && @srid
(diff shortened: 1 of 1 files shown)


mjy:
Order doesn't matter in this context, does it?

```ruby
3.3.4 :009 > def foo(a,b, c: nil, d: 9,  **)
3.3.4 :010 >   puts a,b,c,d
3.3.4 :011 > end
 => :foo 
3.3.4 :012 > foo(1,2,d:22, c: 1)
1
2
1
22
 => nil 
```

vicocamacho:
> Order doesn't matter in this context, does it?

No, it shouldn’t, you are correct. Then the underlying issue should be something else. The weird thing is that I have an internal CI that was having the errors reported on the issue referenced here and they stopped happening once I applied this patch.

I’ll  keep investigating

vicocamacho:
nvm, this is not fixing the issue, closing for now. I will keep investigating.

Issue #378, opened 2023-05-16
Some ActiveRecord Tests Failing


keithdoggett:
Running `bundle exec rake test:activerecord` yields ~50 test failures, currently. Most of these need to be excluded and re-written in our own test suite. Once I have a standard process for excluding ActiveRecord tests I'll update this issue.

BuonOmo:
@keithdoggett I'm back on running this test suite, and excluding tests with `minitest-exclude`. But as of now something quite weird happens: in CI (not locally) tests are super slow when `use_transactional_tests = false`, and they seem to get exponentially slower in a given class. Example from one of my latest CI tests:

```
PostgresqlHstoreTest#test_array_cycle = 0.07 s = .
PostgresqlHstoreTest#test_array_strings_with_array_delimiters = 0.08 s = .
PostgresqlHstoreTest#test_array_strings_with_commas = 0.09 s = .
PostgresqlHstoreTest#test_array_strings_with_null_strings = 0.15 s = .
PostgresqlHstoreTest#test_array_strings_with_quotes = 0.42 s = .
PostgresqlHstoreTest#test_arrow = 1.40 s = .
PostgresqlHstoreTest#test_backslash = 5.20 s = .
PostgresqlHstoreTest#test_cast_value_on_write = 20.81 s = .
PostgresqlHstoreTest#test_change_table_supports_hstore = 82.67 s = .
PostgresqlHstoreTest#test_changes_in_place = 333.01 s = .
... (last is still running!)
```

Any hint on this? Seems fun to debug anyway, the only issue is that it is CI specific.

BuonOmo:
I've added the test suite and excluded tests failing with the `TRIAGE_MESSAGE` constant (`rg TRIAGE MESSAGE` yields them all). Anyone can take some of these once #405 is merged, they are now truly a good first issue.

Note that some _skipped_ tests should not be skipped as well. We should mention them somewhere in our backlog..

keithdoggett:
@BuonOmo agree about creating a backlog for those tests. Could probably make an issue for each one with some kind of new label. 

For the slow tests is there any action we can take? Seems like an upstream issue. Thanks for digging into that.

BuonOmo:
> Could probably make an issue for each one with some kind of new label.

I wouldn't as it would clutter the backlog a lot.

> For the slow tests is there any action we can take? Seems like an upstream issue. 

[...]

PR #405, opened 2024-09-04
Bump to 7.2.0

### LINKED ISSUE #395: Inquiry on Rails 7.2 Support Effort

BuonOmo:
Closes #403 
Closes #395


lib/active_record/connection_adapters/postgis_adapter.rb
```diff
@@ -17,7 +17,6 @@
 require_relative "postgis/oid/spatial"
 require_relative "postgis/oid/date_time"
 require_relative "postgis/type" # has to be after oid/*
-require_relative "postgis/create_connection"
```
BuonOmo, line 20: This is the only actual needed change according to the current test suite, but I'm trying with the whole rails test suite to be sure we do not miss anything

activerecord-postgis-adapter.gemspec
```diff
@@ -10,16 +10,17 @@ Gem::Specification.new do |spec|
 
   spec.version = ActiveRecord::ConnectionAdapters::PostGIS::VERSION
   spec.authors = ["Daniel Azuma", "Tee Parham"]
-  spec.email = ["dazuma@gmail.com", "parhameter@gmail.com", "kfdoggett@gmail.com"]
+  spec.email = ["kfdoggett@gmail.com", "buonomo.ulysse@gmail.com"]
```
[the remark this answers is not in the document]
seuros: Sure.

Rakefile
```diff
@@ -3,7 +3,7 @@ require "rake/testtask"
 require_relative "test/rake_helper"
 
 task default: [:test]
-task test: "test:postgis"
+task test: "test:all"
```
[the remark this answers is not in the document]
BuonOmo: I'll look at it. I have good experience with this suite now that I maintain the CRDB adapter. It usually is flaky due to some tests modifying the db. Which they should not do if they are correctly changed for our use case. Do you have any further information?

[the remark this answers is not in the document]
seuros: Just this : https://github.com/rgeo/activerecord-postgis-adapter/issues/378

activerecord-postgis-adapter.gemspec
```diff
@@ -10,16 +10,17 @@ Gem::Specification.new do |spec|
 
   spec.version = ActiveRecord::ConnectionAdapters::PostGIS::VERSION
   spec.authors = ["Daniel Azuma", "Tee Parham"]
```
seuros, line 12: name should be updated and index matched with the emails.

test/cases/attributes_test.rb
```diff
@@ -99,20 +99,20 @@ def test_joined_spatial_attribute
     private
 
     def create_foo
-      Foo.connection.create_table :foos, force: true do |t|
+      Foo.lease_connection.create_table :foos, force: true do |t|
```
seuros, line 102: these test case work in isolation

activerecord-postgis-adapter.gemspec
BuonOmo: these are authors, I do not consider myself as author :/

test/cases/attributes_test.rb
BuonOmo: What do you mean?

Rakefile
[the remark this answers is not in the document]
BuonOmo: Thx, I plan on excluding using `minitest-exclude` and opening an issue for excluded tests

activerecord-postgis-adapter.gemspec
seuros: Normally, you keep the old authors and maintainers and prepend active ones.

test/cases/attributes_test.rb
BuonOmo: Thanks for act, I still don't get your first point though..

activerecord-postgis-adapter.gemspec
BuonOmo: Not according to the specification reference:

> **AUTHORS=(value)
>
> A list of authors for this gem.
> 
> [...]
>
> **Email**
>
> A contact email address (or addresses) for this gem

<sup>

https://guides.rubygems.org/specification-reference/

</sup>

I'd rather not mention myself as author unless I rewrite a significant portion of the gem alone.

Rakefile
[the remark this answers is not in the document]
keithdoggett: @BuonOmo yeah I hadn't figured out how to replicate the exclusions like in the CRDB adapter, but we definitely need that here since some are failing (which is expected). But to @seuros 's point that's why I just left the postgis tests for now.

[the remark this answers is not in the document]
BuonOmo: Now we only have Postgis basic tests flaky. If one of you has time to look into it it would be much appreciated. Otherwise maybe we can skip them for now and make this or top priority issue, just so we can release this ?

vfonic:
Hey @BuonOmo! Thanks for the great work you've done so far on support for Rails 7.2!

I've tried switching to this branch and upgrading to Rails 7.2.

Here's the error I'm getting when I try to load my ActiveAdmin dashboard page, I don't know how to take it from here. I hope it's useful to you:
Screenshot 2024-09-10 at 14 25 49

Thanks!

vfonic:
Thanks @BuonOmo, I'll have a look.

What do you mean by
> replace the activerecord adapter with postgresql and see if you still have the failure.
?

Thanks!

Rakefile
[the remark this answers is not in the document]
keithdoggett: @BuonOmo so just to make sure I'm clear, there's a few `postgis` tests that were working and are now flakey?  I can try to work through some of them. Is it only in the CI?

BuonOmo:
@vfonic in your codebase, stop using `active-record-postgis-adapter`, and in your database configuration file (likely `database.yml`) set `postgres` instead of `postgis`. And check if this error still occurs.

januszm:
I've noticed that after switching to github version of the gem, although it works with Rails 7.2, it's a bit picky about Postgis setup. I have a remote database configured in the app:

```ruby
class Listing < ApplicationRecord
  establish_connection :data_warehouse unless Rails.env.test?
  self.table_name = 'raw_data.listings'
```

that I completely mock in tests with:

```ruby
allow(Listing).to receive(:where).with(anything).and_return(fake_data)
```

but the test, after the upgrade of this gem, fails with:
```ruby
allow(Listing).to receive(:where).with(anything).and_return(listings)
```

```ruby
     ActiveRecord::StatementInvalid:
       PG::InvalidSchemaName: ERROR:  schema "raw_data" does not exist
       LINE 10:  WHERE a.attrelid = '"raw_data"."listings"'::...
```

So it somehow tries to verify the existence of all schema despite some being mocked. And this didn't happen in earlier versions.

BuonOmo:
I'm not sure I can have the test suite fixed tonigh as I spent already all my afternoon on it, so here are some notes for someone (might be me, later) that want to pick that up.

The test suite gets corrupted at some point. Some setup change (likely the `spatial_factory_store`) and it is not properly reset. The whole challenge is to find which test is the culprit, fix it. The second step would be to avoid a potential reproduction of this issue (I left a TODO addressing this in the code, only useful if my assumptions were right).

If I were to continue on this, I would replace the `teardown` block added with something more precise that directly detects the state corruption.

mjy:
Can you store the identify (`object_id`) of the model (go Ruby!) on spinup and compare it on teardown as a quick way to check if it has changed?

BuonOmo:
@mjy it does change actually, we reinstanciate a new factory pretty often ! So the object id is not enough unfortunately :/

formigarafa:
Made some attempts focusing on `test:postgis`, the flakiness seems related to order the tests run. I am not sure if the same SEED param will make the results deterministic across architectures, but these are these are some repeatable results I got by fixing the seed param.

```
$ SEED=170 ARCONN=postgis bundle exec rake test:postgis
66 runs, 223 assertions, 0 failures, 0 errors, 0 skips
```

```
$ SEED=1 ARCONN=postgis bundle exec rake test:postgis
66 runs, 206 assertions, 3 failures, 1 errors, 0 skips
```

```
$ SEED=2 ARCONN=postgis bundle exec rake test:postgis
66 runs, 218 assertions, 4 failures, 0 errors, 0 skips
```

formigarafa:
I've got to something more specific:
`BasicTest#test_save_and_load_point` fails when it runs after `BasicTest#test_spatial_factory_attrs_parsing`.

Rakefile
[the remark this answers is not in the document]
formigarafa: I noticed some of the tests failures are order related.
notes here: https://github.com/rgeo/activerecord-postgis-adapter/pull/405#issuecomment-2409240511

formigarafa:
Please have a look at #412 . Postgis tests are looking better, now.

test/cases/ddl_test.rb
```diff
@@ -270,26 +273,29 @@ def test_create_spatial_column_default_value_geographic
...
+      # `SpatialColumnInfo#all` queries column info from the database.
+      # It should not be called when klass.columns is called
+      assert_queries_count(0) do
+        # first column is id, second is name
+        refute klass.columns[1].spatial?
+        assert_nil klass.columns[1].has_z
+      end
     end
 
     # Ensure that null contraints info is getting captured like the
```
tagliala, line 295: is it worth to fix the typo here?

```suggestion
    # Ensure that null constraints info is getting captured like the
```

test/test_helper.rb
```diff
@@ -84,6 +107,11 @@ def geographic_factory
       RGeo::Geographic.spherical_factory(srid: 4326)
     end
 
+    # TODO: rather than using this, we should somehow make this a
+    #   fixture that has its self-handled lifecycle. Right now we
+    #   are depending on the dev running `reset_spatial_store` if
+    #   they made any update. They can forget it (with the current
+    #   flacky tests we have, it seems that is actually the case).
```
tagliala, line 114:
```suggestion
    #   flaky tests we have, it seems that is actually the case).
```

tagliala (commented):
Hello, thanks for this work here

I randomly ran into a couple of typos, one is not related to this PR but there was a change close enough

seuros (commented):
Should we keep the excludes files ?

BuonOmo:
@seuros I'm not sure I get you. Yes we should keep the files I added, they correspond to failing tests! And let's track them in the related issue #378.

formigarafa:
About these ignore files: are they working as temporary to reduce the noise while this work is done (like a skip), are they mostly definitive like for unnecessary tests or tests that do not apply in the context of PostGIS or are they a mix of both?

I am asking because I've made a few adjustments to make the ignored test appear as skipped to have an idea of the size of the problem and I wouldn't mind sharing that if it appears useful to get over of what is missing.

BuonOmo:
@formigarafa I think they would be a mix of both. I haven't checked them individually besides just seeing that they were failing.

> I am asking because I've made a few adjustments to make the ignored test appear as skipped to have an idea of the size of the problem and I wouldn't mind sharing that if it appears useful to get over of what is missing.

This looks like a change for the gem responsible of ignoring tests ? Anyway, feel free to submit this!

keithdoggett (approved):
Thank you for the work on this everyone! @BuonOmo would you like me to handle creating a stable branch for the old version and bumping/releasing this?

BuonOmo:
@keithdoggett lets go! Thank you :)

I'll take a look at the new PR #415  soon, but I guess we could already push this, and have a patch release for rake task fix

seuros (approved):

keithdoggett:
v10.0.0 is released. `9.0-stable` branch also up if anyone wants to build directly from git for AR7.1. Thanks everyone!

PR #412, opened 2024-10-14
Fix tests of #405 "Bump to 7.2.0"

formigarafa:
Fix tests setup and reset of #405:
Load current default spatial factory value as is instead of making rgeo produce one. Default factory produced by rgeo is not memoized by rgeo, but the set one is.
Resetting the default to a produced was fixing it to one fixed value.

Revert to use ActiveSupport::TestCase instead of ActiveRecord::TestCase on dd_test, it was causing the tests to attempt loading an misconfigured db connection.

### DIFF
--- a/test/cases/basic_test.rb
+++ b/test/cases/basic_test.rb
@@ -146,7 +146,7 @@ module PostGIS
                                                 has_z: false, has_m: false })
 
       # wrong factory for default
-      old_default = spatial_factory_store.default
+      old_default = spatial_factory_store.instance_variable_get :@default
       spatial_factory_store.default = RGeo::Geographic.spherical_factory(srid: 4326)
 
       object = klass.new
--- a/test/cases/basic_test.rb
+++ b/test/cases/basic_test.rb
@@ -18,7 +18,8 @@ module PostGIS
 
     def test_postgis_available
       assert_equal "PostGIS", SpatialModel.lease_connection.adapter_name
-      assert_equal SpatialModel.lease_connection.select_value("SELECT postgis_lib_version()"), SpatialModel.lease_connection.postgis_lib_version
+      expected_postgis_lib_version_value = SpatialModel.lease_connection.select_value("SELECT postgis_lib_version()")
+      assert_equal expected_postgis_lib_version_value, SpatialModel.lease_connection.postgis_lib_version
       valid_version = ["2.", "3."].any? { |major_ver| SpatialModel.lease_connection.postgis_lib_version.start_with? major_ver }
       assert valid_version
     end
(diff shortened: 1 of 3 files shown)


test/cases/basic_test.rb
```diff
@@ -146,7 +147,7 @@ def test_spatial_factory_attrs_parsing
                                                 has_z: false, has_m: false })
 
       # wrong factory for default
-      old_default = spatial_factory_store.default
+      old_default = spatial_factory_store.instance_variable_get :@default
```
BuonOmo, line 150: I'm not sure this is an ok behaviour in the lib (I mean if it is not clear when testing it, it may not be clear when using it), you've just work on it @formigarafa WDYT?

test/cases/schema_statements_test.rb
```diff
@@ -6,7 +6,7 @@ module PostGIS
   class SchemaStatementsTest < ActiveSupport::TestCase
     def test_initialize_type_map
       SpatialModel.with_connection do |connection|
-        assert connection.connected?
+        connection.connect!
```
BuonOmo, line 9: I guess the reasoning here is that `connect!` would raise if not able to connect? Do you have any idea why this has changed though?

BuonOmo (commented):
Thank you for the fixes! It looks promising 💯 

I have a few questions that I would like to clarify to be sure we're heading in the good direction. Nothing blocking though.

BuonOmo (approved):
Thinking about how this is wanted by the community, let's merge. I'll open an issue in which we'll discuss how to make these tests overall more reliable (thinking using fixtures for factories for instance). I'm still very interested in answering those questions though!

@keithdoggett all good for you too?

test/cases/basic_test.rb
formigarafa: I did not like that either but there is no other way to infer if there is a default set in the factory store.

From `rgeo-ativerecord`: https://github.com/rgeo/rgeo-activerecord/blob/master/lib/rgeo/active_record/spatial_factory_store.rb#L20-L22
```ruby
      def default(attrs = {})
        @default || default_for_attrs(attrs)
      end
```
Maybe If this method used `||=` then it would work as the code seem to expect.

test/cases/schema_statements_test.rb
formigarafa: This was added in #397 , and there is a comment about it: https://github.com/rgeo/activerecord-postgis-adapter/pull/397/files#r1646652915

I think this line was added to highlight why the next command was failing, maybe to try pinpoint the reason the connection sometimes is not there.
I found that this is also dependent of the order the tests run. If you run this test in isolation it would aways fail. If this test run after a test using the connection (eg: `model.connection.create_table`), then this test would pass.

test/cases/basic_test.rb
BuonOmo: I'm unfortunately not sure it is that simple, for instance with `||=` you'd get a weird behaviour doing:

```ruby
f1 = store.default(some_attr)
f2 = stor.default(other_attr)
```

both would be the same, ignoring `other_attr`.

I opened an issue to see if there is a good way to refactor this https://github.com/rgeo/rgeo-activerecord/issues/80

PR #417, opened 2024-11-05
apply quoting patch

JamesChevalier:
From https://github.com/rgeo/activerecord-postgis-adapter/issues/402#issuecomment-2371701919 in https://github.com/rgeo/activerecord-postgis-adapter/issues/402

Includes a test that specifically verifies this query syntax

### DIFF
--- a/lib/active_record/connection_adapters/postgis/quoting.rb
+++ b/lib/active_record/connection_adapters/postgis/quoting.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module ActiveRecord
+  module ConnectionAdapters
+    module PostGIS
+      module Quoting
+        def type_cast(value)
+          case value
+          when RGeo::Feature::Instance
+            value.to_s
+          else
+            super
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


keithdoggett (approved):
Thanks @JamesChevalier! LGTM. I'll see if @BuonOmo or @seuros want to review as well but will merge in a day or 2 if we don't hear from them

BuonOmo (approved):

seuros (approved):

lib/active_record/connection_adapters/postgis/quoting.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module ActiveRecord
+  module ConnectionAdapters
+    module PostGIS
+      module Quoting
+        def type_cast(value)
+          case value
+          when RGeo::Feature::Instance
+            value.to_s
```
pedros007, line 10: After writing https://github.com/rgeo/activerecord-postgis-adapter/issues/402#issuecomment-3071348975 I think perhaps this needs to be extended to support `RGeo::Geos` geometries.  I am unable to do this now but documenting here in case it helps someone in the future.

PR #418, opened 2024-11-05
Fix loading issue with rake db:create on Rails 7.2

### LINKED ISSUE #415: db:create not working on branch bump-7-2
When I try running `rake db:create`, I get the following error:

```
ActiveRecord::Tasks::DatabaseNotSupported: Rake tasks not supported by 'postgis' adapter (ActiveRecord::Tasks::DatabaseNotSupported)
```

If I add this in an initializer, it works:

`ActiveRecord::Tasks::DatabaseTasks.register_task(/postgis/, "ActiveRecord::Tasks::PostgreSQLDatabaseTasks")`

Not sure if this is something that used to be handled automatically and no longer is with Rails 7.2. I can whip up a PR if y'all like, just wanted to double check if I was doing something wrong.

dorner:
Fixes #415.

This issue seems to only happen if a connection isn't created during startup. Not sure why other representative apps don't have this problem - there might be something in an initializer that creates the connection before trying to run the rake task. But from a pure perspective, the postgis adapter file may not load by the time the db:create tasks runs.

This PR moves the registration to the topmost file, which seems to solve this problem.

### DIFF
--- a/lib/active_record/connection_adapters/postgis_adapter.rb
+++ b/lib/active_record/connection_adapters/postgis_adapter.rb
@@ -161,5 +161,4 @@ module ActiveRecord
     spatial_ref_sys
     topology
   ]
-  Tasks::DatabaseTasks.register_task(/postgis/, "ActiveRecord::Tasks::PostgreSQLDatabaseTasks")
 end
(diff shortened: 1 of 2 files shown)


BuonOmo:
I saw in the issue that you had issues testing it. Could you test it manually at least ? And did it work properly ?

seuros:
I could not reproduce this.

dorner:
Let me try to put together a minimal app that can reproduce it.

dorner:
Will post back at the issue.

dorner:
> I saw in the issue that you had issues testing it. Could you test it manually at least ? And did it work properly ?

Yep, I tested this locally and confirmed that it doesn't work on the latest (10.0) but does work on my branch.

keithdoggett (approved):
I was also unable to reproduce this, but as long as this doesn't break anything else I don't have an issue with moving this to load earlier

dorner:
@keithdoggett did you check what I posted on the issue? I was able to reproduce it with a multi-database configuration.

keithdoggett:
Great thanks @dorner

PR #422, opened 2025-01-08
Fix AR type cast to mirror quoting

nikolai-b:
It seems wrong that

```ruby
SpatialModel.where("ST_DWITHIN(latlon_geo, ?, 500)", geographic_factory.point(-72.099, 42.099)).to_sql
```

uses the EWKB format but when the query is actually executed it uses `to_s` which uses the WKT without passing any SRID.  This was causing me issues for me with preprepared statements but I couldn't create a simple text.  I do think this gem should pass the SRID of bindings though.

I'm not very happy with the test I've written here.  I looked `.values_for_queries` or looking at the arel but the binds haven't been type_cast at this point so I've used explain.

### DIFF
--- a/lib/active_record/connection_adapters/postgis/quoting.rb
+++ b/lib/active_record/connection_adapters/postgis/quoting.rb
@@ -5,8 +5,9 @@ module ActiveRecord
     module PostGIS
       module Quoting
         def type_cast(value)
-          case value
-          when RGeo::Feature::Instance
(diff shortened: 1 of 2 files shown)


keithdoggett (commented):
@nikolai-b thanks for the PR! I think this looks good. Was this only effecting prepared geometries? I thought there was some work done recently that added the SRID to the generated string in the rgeo-activerecord gem.

PR #423, opened 2025-01-28
Added back Ruby 2.7 support for 9.0-stable (AR 7.1).

Jack12816:
**- What is it good for**

Unfortunately the support for Ruby 2.7 on the 9.x branch which supports ActiveRecord 7.1 was cut off. The required minimum Ruby version is 3.0. Rails (including AR) on the other hand [supports Ruby 2.7](https://github.com/rails/rails/blob/7-1-stable/rails.gemspec#L12), without issues (also tested via BuildKite). This patch brings back the Ruby 2.7 support for the 9.x branch in order to be compatible with ActiveRecord 7.1. The issue occured for me while upgrading an application from Rails 6.1 to 7.1 (the last version with Ruby 2.7 support).

**- What I did**

Just re-enabled the Ruby 2.7 support on CI and lowered the minimum required Ruby version on the gemspec, and the Rubocop config. I've already created a testing branch with Ruby 2.7-3.4 to verify the code base does not need any modifications, [tested via CI](https://github.com/Jack12816/activerecord-postgis-adapter/actions/runs/13006623590), [2nd test](https://github.com/Jack12816/activerecord-postgis-adapter/actions/runs/13006526472/job/36274521279).

**- A picture of a cute animal (not mandatory but encouraged)**

image

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ruby-head, '3.2', '3.1', '3.0']
+        ruby: [ruby-head, '3.2', '3.1', '3.0', '2.7']
         pg: [11-3.0, 12-master, 13-master, 14-master, 15-master]
     steps:
       - name: Set Up Actions
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,7 +1,7 @@
 inherit_from: .rubocop_todo.yml
 
 AllCops:
-  TargetRubyVersion: 3.0
+  TargetRubyVersion: 2.7
   DisplayCopNames: true
   SuggestExtensions: false
   NewCops: enable
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ _JRuby support for Rails 4.0 and 4.1 was added in version 2.2.0_
 
 ```
 ActiveRecord 7.1
-Ruby 3.0.0+
+Ruby 2.7.0+
 PostGIS 2.0+
 ```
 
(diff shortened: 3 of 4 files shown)


BuonOmo:
Hi! Thank you for the contribution. Ruby 2.7 was EOL [in March 2023](https://www.ruby-lang.org/en/downloads/branches/). As we already are having a hard time maintaining for current rails and ruby versions, I'll close this. @keithdoggett feel free to reopen if you believe I am wrong :)

P.S: is there a cute animal picture API ? I'm interested

Jack12816:
Mhh :( I get the point, but the PR is small and targeting the same dependencies as Rails for a monkey-patch gem seems reasonable to me. Even more when its *just config*. There is no actual change to the code base. 

At least working around this is simple: just adding a git reference to my fork branch on the projects Gemfile which is going to be migrated. (I anybody gets hit by the same issue) 

```ruby
gem 'activerecord-postgis-adapter',
    github: 'Jack12816/activerecord-postgis-adapter',
    branch: '9.0-stable'
```

---

> P.S: is there a cute animal picture API ? I'm interested

Just google "cute animal", select and copy a little image :)

BuonOmo:
@Jack12816 the snippet you gave is perfect. The point is that I don't want people to believe we support Ruby 2.7. We don't and we won't backport any fix. Hence better that we show it openly by not allowing it :)

If someone want to bypass that, it is at their own risks, and copy-pasting the snippet you gave shows it clearly!

mjy:
I agree @BuonOmo, this also prevents downstream technical debt that might emerge- "Why are tests failing now?".

PR #433, opened 2025-07-24
Fix: don't match enum columns as spatial

modosc:


BuonOmo:
Hey @modosc, could you update your PR and add an entry to the History.md file please ? I also am not 100% sure about how this is handled yet. May I work on it as well or would you rather keep ownership ? If I may, I'd like access to your repo as a contributor! In any case, I'll work on the activerecord-cockroachdb-adapter version of it now hence we'll be able to compare in the end :)

modosc:
@BuonOmo i just rebased it. i can't give you access because it's originating from my work org.

BuonOmo:
Ok so what we need:

- [ ] History entry
- [ ] Correct test lifecycle (EDIT: we clearly see how test lifecycle is messed up in tests results)

What I'm unsure about:

- [ ] the place where the fix is made (we're already setting geometry related instance variables beforehand, do we want to? also the complexity of the method is becoming huge)

Let's work on that together then :)

PR #435, opened 2025-10-03
Compatibility with Rails 8.1.beta1

t27duck:
- Handle `cast_type` being the second parameter for `SpatialColumn.initialize`.
- Handle nil case during OID initiation.
- Handle case where in `spatial_factory` the object could be frozen for whatever reason.
- Update CI to test against 8.1 beta and supported postgresql versions.

**These changes require the changes from https://github.com/rgeo/rgeo-activerecord/pull/83 so that gem is set to pull from the repo in this branch.**

At least locally, there are 5 failures and 2 errors in the test suite with these changes. The failures are all in the deprecation tests and the errors are related to connection tests (I think).

I've been running both this branch and the changes to `rgeo-activerecord` in an app for several weeks with no issue, though granted the app only really cares about storing and retrieving geometry data with minimal to no querying.

If nothing else, here's a good starting point.


lib/active_record/connection_adapters/postgis/oid/spatial.rb
```diff
@@ -25,6 +26,11 @@ def initialize(geo_type: "geometry", srid: 0, has_z: false, has_m: false, geogra
           #   has_z:    false
           #   has_m:    false
           def self.parse_sql_type(sql_type)
+            # Could be nil during type registration
```
t27duck, line 29: Not entirely sure why this happens as I'm not fully familiar with active record internals, but this seems to work.

seuros:
There are more changed required to have the adapter working.  

For this reason, i decided to create an [extension](https://github.com/seuros/activerecord-postgis) instead of legacy way of having a pg adapter..

My goal is to move that repo later to this organization. The interface is plug & play.

t27duck:
Actually looking at the failed tests, they seem to all be from the ActiveRecord test suite itself. Is there a reason why the gem runs the tests for Rails?

Gemfile
```diff
@@ -6,6 +6,8 @@ gemspec
 gem "pg", "~> 1.0", platform: :ruby
 gem "byebug" if ENV["BYEBUG"]
 
+gem "rgeo-activerecord", git: "https://github.com/rgeo/rgeo-activerecord.git"
```
seuros: interesting!

Do rgeo need any release ?

t27duck:
Current state of things:

Locally, tests _should_ pass now with the inclusion of those backtrace fixes.

I removed the `frozen?` check in `#spatial_factory`.

I tried moving that logic into the initializer, but ran into two issues:
- `geo_type` could be `nil` (probably a result of that nil check in `.parse_sql_type`.
- After casting `geo_type` as a string always, the remaining failing test is ActiveRecord checking if the object is safe for ractors, which it wasn't with the change.

t27duck:
Looks like the runners that failed all failed on the same error.

```
  1) Error:
CounterCacheTest#test_reset_counters_for_multiple_records:
ActiveRecord::Fixture::FixtureError: table "dogs" has no columns named "trainer_id", "dog_lover_id".
```

Which is an ActiveRecord test for fixtures. Possibly a flakey test?

gap777:
I'm a little confused... @t27duck is the path forward to adopt the new gem that @seuros is proposing (whereby the pg adapter is extended rather than replaced), or to bring in Rails 8.1 compatibility with this gem? 

We're blocked on our upgrade and would love assistance/guidance on where to go with this issue.

BuonOmo:
@gap777 

> whereby the pg adapter is extended rather than replaced

This plan has not been discussed by maintainers of the gem, you'd have to ask @seuros. As far as I'm concern, I'd rather keep this gem.

I'm currently working on rails compatibility for the CRDB adapter, once done I'll have time to handle this more properly. In the mean time, you can point to this branch. The differences are usually in some deep functionnalities and you're unlikely to have problems. And if you do, please report them here.

> We're blocked on our upgrade and would love assistance/guidance on where to go with this issue.

You can talk to your company about RGeo's open-collective and financing. Or directly hire me as a freelance if you need fast-paced guidance and proper deadlines. Otherwise, as we're few to maintain on our spare time, I cannot give you deadlines.

BuonOmo:
> Yes, the email was correct, but I suspect it went to spam. Keith did receive it.

I just checked again for the keywords `"seuros"`, `"activerecord"`, and `"postgis"` within my whole mailbox + spam, nothing :/

> My only concern with solo maintenance is that it might work fine for my use case but break for others.

My other concern would also be that the day you decide to stop working on it, the whole community loses the gem... At least here we're two maintainers, and a few active developers.

> I can maintain it, it’s smaller, cleaner, and has more tests.

Could you maybe make an issue in this repo explaining the main axes of differences and how we could refactor to get to an ideal place that keeps steps made here, history and still take advantage of the new rails architecture?

> Checked locally, the failure are just timing related. :shipit:

What do you mean? I believe there is still the ractor issue that we need to either tackle or create an issue for. And about the failing tests, if by timing you mean that they depend on test order, that is still not green IMHO. I've been using `minitest-bisect` with the crdb adapter for that purpose. I'll be able to do it here as well once I've shipped the crdb adapter. Otherwise, feel free to do it :)

seuros:
> My other concern would also be that the day you decide to stop working on it, the whole community loses the gem... At least here we're two maintainers, and a few active developers.

That exactly why i asked it to move it to  `https://github.com/rgeo/activerecord-postgis` instead. Was not asking  to have traffic send to my repo while keeping the gem.

----

I use the gem heavily, i did not see any regression.

BuonOmo:
> I use the gem heavily, I did not see any regression.

Then users can point to this branch with some sense of security. I would still focus on fixing tests and make sure ractor compatibility is correct. I would be okay with having an `8-1` branch that people can point to until we are ready for a release, if any interest?

> That exactly why I asked it to move it to https://github.com/rgeo/activerecord-postgis instead. Was not asking to have traffic send to my repo while keeping the gem.

Well, I'm up to opening PRs here to migrate some parts of your code (for instance the arel part, but I'm sure you know of others that would make sense). I still am against patching pg adapter and think we should rather inherit (as mentioned above), but that could be up for discussion.

t27duck:
@BuonOmo feel free to fork my fork if you want. 👍

BuonOmo:
@seuros here is what I did:

```
gh co StreetIQ:allow_rails81
# commit

# And either
git push # fatal: The upstream branch of your current branch does not match
git push origin HEAD:refs/pull/435/head # deny updating a hidden ref
git push origin HEAD # [new branch]      HEAD -> allow_rails81
```

I also tried to add the StreetIQ repo to my remotes and push there but it was refused (hence my belief that I had not been allowed

Any other lead ?

Maybe I should have run `gh co 435` rather than the branch name?

t27duck:
I'm fine with whatever. My app can coast on this branch if being able to do your own thing is easier.

BuonOmo:
I made a new PR, please review it, I'll ship soon if nothing is blocking!

PR #437, opened 2025-11-27
fix: missing @srid instance variable in v11.1 causes breaking change

oleksii-leonov:
`def wkt_parser(string)` in `ActiveRecord::ConnectionAdapters::PostGIS::OID::Spatial` internally uses `@srid`:

```ruby
          def wkt_parser(string)
            if binary_string?(string)
              RGeo::WKRep::WKBParser.new(spatial_factory, support_ewkb: true, default_srid: @srid)
            else
              RGeo::WKRep::WKTParser.new(spatial_factory, support_ewkt: true, default_srid: @srid)
            end
          end
```

In 11.0 to 11.1 change, due to initializer refactoring, the `@srid` instance variable was removed from the initializer, but `def wkt_parser(string)` still uses it (so it's always `nil`, which breaks things in some situations).

Adding `@srid` back as a hotfix for the breaking change.

### DIFF
--- a/lib/active_record/connection_adapters/postgis/oid/spatial.rb
+++ b/lib/active_record/connection_adapters/postgis/oid/spatial.rb
@@ -12,6 +12,7 @@ module ActiveRecord
           def initialize(geo_type: "geometry", srid: 0, has_z: false, has_m: false, geographic: false)
             super()
             @geographic = geographic.freeze
+            @srid = srid.freeze
             @factory_attrs = {
               geo_type: geo_type.underscore.freeze,
               has_m: has_m.freeze,


BuonOmo:
@oleksii-leonov Your change doesn't take into account the `hash` and `==` methods. Seeing that this should be deployed quickly, I'll cherry-pick and use it. Thank you for detecting that.

We should definitely test that method.

BuonOmo:
Published

oleksii-leonov:
@BuonOmo , thank you a lot!

`hash` and `==` using `@geographic, @factory_attrs`, so not affected.

Actually, I am not sure that `default_srid:` in `def wkt_parser(string)` is in use, as `spatial_factory` will always have the same `srid` defined.

- If we pass `srid != nil` in the initializer, spatial_factory will have `srid` defined (and  `default_srid:` will not be used).
- If we pass `srid == nil` in the initializer, then `default_srid: @srid` will be used, but also will be `nil`, so it has no effect.

So we may be able to remove `default_srid:` completely to reduce complexity.

I have caught this problem only because in our project, we have a monkey-patch for `ActiveRecord::ConnectionAdapters::PostGIS::OID::Spatial`, which is also using the `@srid` instance variable (it's only our own problem as we are using `ActiveRecord::ConnectionAdapters::PostGIS::OID::Spatial` private internals, and we already fixed it by using `@factory_attrs[:srid]`).

BuonOmo:
@oleksii-leonov Feel free to open a subsequent PR removing the `default_srid` and adding tests to prove your point and ensure that don't fail again on that!

PR #440, opened 2026-02-10
Fix limit data dropped from schema on migrate

conorml:
Add encode_with/init_with methods to SpatialColumn so that spatial attributes (geographic, geometric_type, has_m, has_z, srid, limit) are properly serialized and deserialized during schema cache operations.

Also implement == and hash so that schema change detection correctly compares spatial columns.

Problem: when running rails db:migrate there is always a diff in the schama.rb file for each table of type t.geography ex:
t.geography "lonlat", limit: {:srid=>4326, :type=>"st_point", :geographic=>true}
=>
t.geography "lonlat"

with this change this diff no longer exists as the column is read fully by the adapter.

NOTE: This has been fixed in master, but is still an issue in 10.x and we're not on rails 8.1 yet so still need to stay on 10.x

### DIFF
--- a/lib/active_record/connection_adapters/postgis/spatial_column.rb
+++ b/lib/active_record/connection_adapters/postgis/spatial_column.rb
@@ -58,6 +58,29 @@ module ActiveRecord  # :nodoc:
           %i[geometry geography].include?(@sql_type_metadata.type)
         end
 
+        SPATIAL_ATTRIBUTES = %w[geographic geometric_type has_m has_z srid limit].freeze
+
+        def init_with(coder)
+          SPATIAL_ATTRIBUTES.each { |attr| instance_variable_set(:"@#{attr}", coder[attr]) }
+          super
+        end
+
+        def encode_with(coder)
+          SPATIAL_ATTRIBUTES.each { |attr| coder[attr] = instance_variable_get(:"@#{attr}") }
+          super
+        end
+
+        def ==(other)
+          other.is_a?(SpatialColumn) &&
+            super &&
+            SPATIAL_ATTRIBUTES.all? { |attr| public_send(attr) == other.public_send(attr) }
+        end
+        alias :eql? :==
+
+        def hash
+          SPATIAL_ATTRIBUTES.reduce(SpatialColumn.hash ^ super.hash) { |h, attr| h ^ public_send(attr).hash }
+        end
+
         private
 
         def set_geometric_type_from_name(name)
--- a/test/cases/spatial_column_test.rb
+++ b/test/cases/spatial_column_test.rb
@@ -1,0 +1,516 @@
+# frozen_string_literal: true
+
+require_relative "../test_helper"
+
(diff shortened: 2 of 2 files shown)


keithdoggett (commented):
I think this looks fine. I'm just trying to figure out the exact issue here. Is the `limit` method not working or is it specifically when loading data in with YAML via `init_with` and `encode_with`?

I'm not super familiar with how that version of the schema generator works so just trying to think through if there's a better way to handle this or if some other method is not working and this is just a patch over top of that. Thanks!

conorml:
> I think this looks fine. I'm just trying to figure out the exact issue here. Is the `limit` method not working or is it specifically when loading data in with YAML via `init_with` and `encode_with`?
> 
> I'm not super familiar with how that version of the schema generator works so just trying to think through if there's a better way to handle this or if some other method is not working and this is just a patch over top of that. Thanks!

It doesn't seem to actually affect the limit method at runtime, but just what is exported to the schema on update.  The way I understand it is that if you clean out your schema and run export the schema it works fine.   The problem comes if you try to update the schema after running the application.  So when activestorage initialized and calls `deduplicate` it drops limit when it updates the schema cache.  And so then when updating the schema file using the schema cache the geography column no longer have the limit data and so the schema for those columns gets updated correctly.  So you can fix this by just clearing the schema cache after running the application and it works fine, but it's just a pain to have to either do that or to make sure to reject these changes to the schema file if they get updated incorrectly.

It's here where the column gets deduplicated and without this change lose the limit property: https://github.com/rails/rails/blob/bb2bdef2925433a0c5db31b873f9faddf2e2e65d/activerecord/lib/active_record/connection_adapters/column.rb#L104

conorml:
@keithdoggett could you re-run the ci tests?  These seem to just be flakey, but want to make sure they can all pass.

keithdoggett:
@conorml looks like the tests are passing and the changes look good given your explanation. I can get this merged and look into deploying a new version for this major version tomorrow.

keithdoggett:
Thanks @conorml merged into `10.0-stable` now. Will post here if a release is made.

andyatkinson:
Hi @conorml @keithdoggett We seem to have a very similar issue where `geography` data type columns will dump the additional `limit` data to `schema.rb` or not. This inconsistency ends up causing tons of noise in migrations PRs as besides the difference to `db/schema.rb` we also use the serialized schema cache dump `db/schema_cache.yml` which ends up with a huge number of changed lines that are unrelated to the original migration.

We're on Rails 7.0.x and so we're limited to 8.x of activerecord-postgis-adapter. I'm not sure on our timeline to move to Rails 7.2 for this fix, so is there any chance this would be relatively easy and low risk to back port to 8.x? We'd be happy to use it and share our experience on whether it solves the issue. We also believe to have a patch on our side to try as what's hopefully a temporary fix before moving to this newer gem version.

Thank you.

PR #441, opened 2026-02-24
Release: 10.0.2

keithdoggett:
Include fix for schema dropping limits on migration

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,7 @@
+### 10.0.2 / 2026-02-24
+
+* Fix limit data dropping from schema on migrate (conormi) #440
+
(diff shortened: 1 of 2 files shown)


keithdoggett:
@BuonOmo @seuros thoughts on making releases for older version like this? I'll make a corresponding update to history on the main branch if this goes through to ensure that has all versions on it.

seuros (approved):
yes , this is not cosmetic change. 

we could back-port if there is a need.

conorml:
@keithdoggett could this be released now and then backported to other versions if needed?

keithdoggett:
Released on Rubygems

